PR #2, opened 2023-05-21
Add support for solution 3000 + polling, and for the alarm being triggered

sanjay900:
This also checks for support of some commands, namely the `PRODUCT_SERIAL` command as this isn't supported on a lot of panels by the looks of things.
Also exposes a list of all triggered alarms as a attribute on the panel.


sanjay900:
We will probably want to work out a nicer way to handle the platform specific stuff as I am currently just using if statements here, and you did say you preference would be to not do that, but im not really sure what a better method is.

bosch_alarm_mode2/panel.py
```diff
@@ -213,49 +296,158 @@ async def _monitor_connection_once(self):
...
-        creds = bytearray(b'\x01')  # automation user
-        creds.extend(map(ord, self._passcode))
-        creds.append(0x00); # null terminate
-        result = await self._connection.send_command(CMD.AUTHENTICATE, creds)
-        if result != b'\x01':
-            self._connection.close()
-            error = ["Not Authorized", "Authorized", "Max Connections"][result[0]]
-            raise PermissionError("Authentication failed: " + error)
-        LOG.debug("Authentication success!")
+        if self.model in ["Solution 2000", "Solution 3000"]:
```
mag1024, line 299: Yeah, the ifs are a bit awkward... Maybe longer term we can think about some kind of a panel-model-specific adaptor that can encapsulate all the stuff that's different per-panel, but for now, let's at least break it up into multiple methods, so we can write something like:

if self.model in ["Solution 2000", "Solution 3000"]:
  self._login_remote_user(...)
else:
  self._authenticate_automation(...)

bosch_alarm_mode2/panel.py
```diff
@@ -213,49 +296,158 @@ async def _monitor_connection_once(self):
...
-            request = bytearray(id.to_bytes(2, 'big'))
-            request.append(0x00)  # primary language
-            request.append(0x01)  # return many
-            data = await self._connection.send_command(name_cmd, request)
-            if not data: break
-            while data:
-                id = _get_int16(data)
-                name, data = data[2:].split(b'\x00', 1)
+        if self._featureCommandRequestAreaTextCF03:
+            while True:
```
mag1024, line 400: Ditto, let's break these out into helper methods

mag1024 (commented):
Oh wow, you've made lots of progress!
Overall, looks good -- just a few comments aiming to clean up the structure a bit.
I'm really happy to see that the polling can just use the loading code and doesn't add a lot of cruft.

Will try to test it on my panel soon.

Thank you again!

bosch_alarm_mode2/panel.py
```diff
@@ -111,6 +130,49 @@ def __init__(self, host, port, passcode):
         self.serial_number = None
         self.areas = {}
         self.points = {}
+        self.area_count = 0
+        self.point_count = 0
```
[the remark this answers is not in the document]
sanjay900: I use these values when populating the entity lists, hence the capacities stuff

bosch_alarm_mode2/panel.py
```diff
@@ -111,6 +130,49 @@ def __init__(self, host, port, passcode):
...
+        self.output_count = 0
+        self.user_count = 0
+        self.keypad_count = 0
+        self.door_count = 0
+        self._part_arming_type = AREA_ARMING_PERIMETER_DELAY
+        self._all_arming_type = AREA_ARMING_MASTER_DELAY
+        self._featureProtocol01 = False
```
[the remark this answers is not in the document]
sanjay900: We likely don't need most of these, I had just copy pasted aist from elsewhere and was planning on pruning it down once we have everything we need working

bosch_alarm_mode2/panel.py
```diff
@@ -69,10 +84,13 @@ def is_part_armed(self):
         return self.status in AREA_STATUS_PART_ARMED
     def is_all_armed(self):
         return self.status in AREA_STATUS_ALL_ARMED
+    def is_triggered(self):
+        return (self.is_part_armed() or self.is_all_armed()) and ALARM_MEMORY_PRIORITY_BURGLARY_ALARM in self.alarms
```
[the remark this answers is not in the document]
sanjay900: Good point, I'll add that in

bin/alarm.py
```diff
@@ -19,7 +19,7 @@
                     level = logging.DEBUG)
 
 loop = asyncio.new_event_loop()
-panel = Panel(host=args.host, port=args.port, passcode=args.passcode)
+panel = Panel(host=args.host, port=args.port, passcode=args.passcode, legacymode=False)
```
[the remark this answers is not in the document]
sanjay900: It's only really there if someone is on such an old firmware, that their firmware doesn't support the commands for deriving, though in saying that, maybe instead we just catch the nak from that and jump into a legacy mode that way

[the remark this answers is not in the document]
sanjay900: thinking about it now, i think those old panels will just not return any data for what is supported, so we will automatically fall back to the older formats in those cases

sanjay900:
I was also very glad when the loading code just worked

bosch_alarm_mode2/panel.py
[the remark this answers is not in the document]
sanjay900: Okay i thought that was necessary, but it turns out it isn't, which is good as some panels dont support the capacities commands anyways. I've now removed it and _load_capacities

bin/alarm.py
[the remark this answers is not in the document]
EHylands: Hi @sanjay900 , always happy to follow up on other projects progress!

With WhatAreYou command format 3, LegacyMode is not as useful on Solution Panel, B Series panel and newer G panels.

On my side, I mostly kept LegacyMode around for older GV4 panels or Amax panels. Only one user tried running it on GV4, but it failed because he was running firmware < 2.0 and didn't want to risk an upgrade.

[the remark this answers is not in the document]
sanjay900: I see, it would be nice to support those older panels. What exactly is the failure you get when you use command format 3?

sanjay900:
For the moment ive pushed the package to pypi just so that i can get people testing a version without needing them to go and manually delete their local git copies when i make changes. I had someone test on the solution 2000 and it is working there, and im currently getting someone to also test on a AMAX300 as well

bin/alarm.py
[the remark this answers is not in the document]
EHylands: As of now, no user reported success or error for these older panels ...

WhatAreYou CF 3 is IIP 5.72, so maybe not supported by older panels or firmware.

With Home Assistant bigger user base, you will probably get request for those panels !

I mostly use LegacyMode now to test the pooling code on my B panel.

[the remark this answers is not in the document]
sanjay900: Ah okay, thank you!

bosch_alarm_mode2/const.py
```diff
@@ -61,6 +61,39 @@
     0xA9: "B6512 (US1B)",
 }
 
+PANEL_AREAS_MAX = {
```
mag1024, line 64: Do we know what panels this is actually necessary for?

In general, to make reviewing easier, I would suggest the approach of focusing this PR on support for the panels we have access to/can directly test with, and splitting additional panel compatibility features into separate PR.

sanjay900:
Yeah, I'm testing with someone's Amax 3000 at the moment

bosch_alarm_mode2/const.py
sanjay900: Currently testing with someone elses AMAX panel, but i was really just trying to see if it would be easy enough to get going without changing too much - though now its starting to look like the answer to that is no, and the AMAX panel is just a bit odd in how it works unfortunately. I will see though, i might be able to get direct access to one.

mag1024: Cool -- but I would suggest doing this kind of exploration on a separate branch so that it's easy to keep track of what's changed, especially in case it doesn't work out. In parallel we can work on addressing the remaining cleanups in this PR to get it merged.

sanjay900: yeah fair enough, i was hoping it would just work but since it doesnt, ill just slice all these commits off

sanjay900: Have now cleaned all this up

sanjay900:
@mag1024 does this all work on your panel?

bosch_alarm_mode2/panel.py
```diff
@@ -211,37 +263,69 @@ async def _monitor_connection_once(self):
...
+            self._all_arming_type = AREA_ARMING_MASTER_DELAY
 
-        data = await self._connection.send_command(
-                CMD.PRODUCT_SERIAL, b'\x00\x00')
-        self.serial_number = int.from_bytes(data[0:6], 'big')
+    async def _basicinfo(self):
+        commandformat = bytearray()
+        commandformat.append(3)
+        
+        data = await self._connection.send_command(CMD.WHAT_ARE_YOU, commandformat)
```
mag1024, line 306: can inline command command format as 'b\x03'

bosch_alarm_mode2/panel.py
```diff
@@ -211,37 +263,69 @@ async def _monitor_connection_once(self):
...
         LOG.debug("Authentication success!")
+    
+    async def _authenticate(self):
+        if self.model in ["Solution 2000", "Solution 3000"]:
+            await self._login_remote_user()
+        else:
+            await self._authenticate_automation()
 
-    async def _basicinfo(self):
+    async def _model(self):
```
mag1024, line 290: Do we need both flavors of WHAT_ARE_YOU?
Can we unify _model and  _basicinfo, and always send format 3?

bosch_alarm_mode2/panel.py
```diff
@@ -256,6 +340,70 @@ async def _load_names(self, name_cmd) -> dict[int, str]:
...
+        index = 0
+        while data:
+            b = data.pop(0)
+            for i in range(8):
+                id = index + (8 - i)
+                if b & 1 != 0:
+                    names[id] = f"{type}{id}"
+                b >>= 1
+            index+=8
+        return names
```
mag1024, line 364: nit: line break between functions

bosch_alarm_mode2/panel.py
```diff
@@ -256,6 +340,70 @@ async def _load_names(self, name_cmd) -> dict[int, str]:
...
+    async def _load_names_cf01(self, name_cmd, names) -> dict[int, str]:
+        for id in names.keys():
+            request = bytearray(id.to_bytes(2, 'big'))
+            request.append(0x00)  # primary language
+            data = await self._connection.send_command(name_cmd, request)
+            name = data.split(b'\x00', 1)[0]
+            names[id] = name.decode('ascii')
+        return names
+
+    async def _load_configured_names(self, config_cmd, type):
```
mag1024, line 352: I think this is a bit mis-named, because it doesn't actually load any *names*.
Maybe something like
_load_configured_entities(self, load_cmd, entity_type) -> dict[int, str]:

bosch_alarm_mode2/panel.py
```diff
@@ -256,6 +340,70 @@ async def _load_names(self, name_cmd) -> dict[int, str]:
...
+        while response_detail:
+            area = _get_int16(response_detail)
+            # item_type = response_detail[2]
+            point = _get_int16(response_detail, 3)
+            if point == 0xFFFF:
+                await self._get_alarms_for_priority(priority, area, point)
+            if area in self.areas:
+                self.areas[area]._set_alarm(priority, True)
+            else:
+                print(f"Found unknown area {area}, supported areas: [{self.areas.keys()}]")
```
mag1024, line 392: Log.warning

bosch_alarm_mode2/panel.py
```diff
@@ -256,6 +340,70 @@ async def _load_names(self, name_cmd) -> dict[int, str]:
...
+
+    async def _get_alarm_status(self):
+        data = await self._connection.send_command(CMD.ALARM_MEMORY_SUMMARY)
+        for priority in ALARM_MEMORY_PRIORITIES.keys():
+            i = (priority - 1) * 2
+            count = _get_int16(data, i)
+            if count:
+                await self._get_alarms_for_priority(priority)
+            else:
+                # If events have been cleared, then we can just clear all areas instead of asking the panel for more info
```
mag1024, line 403: line wrap, please

mag1024 (changes requested):
Apologies for the delay, my HA dev environment has bit-rotted, and I had to re-create it.
Tested it with my panel, and everything seems to still work.

Overall, looks great!
The main comment is about unifying _model and _basicinfo -- everything else is mostly nitpicks.
And thanks again :))

bosch_alarm_mode2/panel.py
```diff
@@ -211,37 +263,69 @@ async def _monitor_connection_once(self):
...
+        creds.append(0x00) # null terminate
         result = await self._connection.send_command(CMD.AUTHENTICATE, creds)
         if result != b'\x01':
             self._connection.close()
             error = ["Not Authorized", "Authorized", "Max Connections"][result[0]]
             raise PermissionError("Authentication failed: " + error)
         LOG.debug("Authentication success!")
+    
+    async def _authenticate(self):
+        if self.model in ["Solution 2000", "Solution 3000"]:
```
mag1024, line 285: BTW, have you tested what happens if we send an AUTHENTICATE to a solutions panel?
Does it reject it as an error, or does it get confused and hang?
If it's the former, maybe we can just blinding try both authentication methods in order all the time without conditioning this on the model and succeed if either succeeds. This would remove the dependency between authenticate and basicinfo.

sanjay900:
One other thing, since not every panel has a serial number, should we be using something else for the unique id?

bosch_alarm_mode2/panel.py
sanjay900: have just changed it now so that it just does both methods

sanjay900:
Im sure theres a better way to handle history - i think tomorrow ill add in checking if the text based history commands are supported for a panel, and ill use that if its available - i know the solution and amax panels dont support it though.
We might also want to look at a more standardized history format, and im not sure if theres a better way to bring that data across to home assistant besides just a giant string in an attribute

mag1024:
The history stuff looks cool, but following the best practice of keeping the changes small/focused, let's please pull it out into a separate PR.

Re: missing serial numbers. I guess we the best we can do at that point is the model? We just need something that is stable and "unique enough" not to have collisions within a given HA instance.

sanjay900:
Fair call, ill splice out the history stuff into its own branch. I was just trying to get something that had the features of my old project, so I can just deprecate that and push people over to this one.

sanjay900:
Out of curiousity though, does that history stuff parse correctly on your panel?

mag1024:
> Out of curiousity though, does that history stuff parse correctly on your panel?

This is what I got:

```
2023-05-25 23:59:19.825 DEBUG (MainThread) [custom_components.bosch_alarm.bosch_alarm_mode2.panel] Loading history...
2023-05-25 23:59:19.825 DEBUG (MainThread) [custom_components.bosch_alarm.bosch_alarm_mode2.connection] >> b'\x01\x06\x15\xff\x00\x00\x00\x00'
2023-05-25 23:59:20.411 DEBUG (MainThread) [custom_components.bosch_alarm.bosch_alarm_mode2.connection] << b'\x01\xde\xfe\t\x00\x00\x0b\x87\x00\x19\x00\x00\x00\x14\x00\x00\x00\x00-\x16\xd9@\x06fffffffff\x000\x00\x00\x00\x00\x00\x00\x00\x00-\x19G\xc0ffffffffff\x00\r\x00\x00\x00\x15\x00\x00\x00\x00-\x19G\xc0\x06fffffffff\x00y\x00\x00\x00\x00\xff\xff\x00\x04-\x19H\x00ffffffffff\x00=\x00\x00\x00\x00\x00\x01\x00\x03-\x1ai\x80ffffffffff\x000\x00\x00\x00\x00\x00\x00\x00\x00-\x1bU@ffffffffff\x00\r\x00\x00\x00\x15\x00\x00\x00\x00-\x1bU@\x06fffffffff\x00y\x00\x00\x00\x00\xff\xff\x00\x04-\x1bU\x80ffffffffff\x00\x13\x00\x00\x00\x04\x00\x00\x00\x00-\x1ci\x00\x06fffffffff'
2023-05-25 23:59:20.411 DEBUG (MainThread) [custom_components.bosch_alarm.bosch_alarm_mode2.panel] Loading count: 9
2023-05-25 23:59:20.411 ERROR (MainThread) [root] Initial panel connection failed
Traceback (most recent call last):
File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/__init__.py", line 38, in async_setup_entry
await panel.connect(load_selector=bosch_alarm_mode2.Panel.LOAD_ALL_HISTORY)
File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/bosch_alarm_mode2/panel.py", line 165, in connect
await self._connect(load_selector)
File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/bosch_alarm_mode2/panel.py", line 241, in _connect
File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/bosch_alarm_mode2/panel.py", line 175, in load
await self._load_history()
File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/bosch_alarm_mode2/panel.py", line 270, in _load_history
area._add_history(self._history_parser(data))
File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/bosch_alarm_mode2/history.py", line 735, in _parse
return b_g_history(date, event_code, area, param1, param2, param3)
TypeError: b_g_history() takes 5 positional arguments but 6 were given
```

bosch_alarm_mode2/panel.py
```diff
@@ -105,19 +136,30 @@ def __init__(self, host, port, passcode):
...
+        self._history = []
 
         self.model = None
         self.protocol_version = None
         self.serial_number = None
         self.areas = {}
         self.points = {}
+        self._partial_arming_id = AREA_ARMING_PERIMETER_DELAY
+        self._all_arming_id = AREA_ARMING_MASTER_DELAY
+        self._supportsSubscriptions = False
```
mag1024, line 152: Let's rename these to lower case with underscores for consistency

bosch_alarm_mode2/panel.py
```diff
@@ -126,12 +188,19 @@ async def load(self, load_selector):
...
-                    self._load_entity_status(CMD.AREA_STATUS, self.areas),
-                    self._load_entity_status(CMD.POINT_STATUS, self.points))
-            await self._subscribe()
+            await self._load_entity_status(CMD.AREA_STATUS, self.areas)
+            await self._load_entity_status(CMD.POINT_STATUS, self.points)
+            if self._featureCommandSetSubscriptionCF01:
+                await self._subscribe()
+            else:
+                loop = asyncio.get_running_loop()
+                self._poll_task = loop.create_task(self._poll())
```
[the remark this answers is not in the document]
mag1024: Actually, we might still have an issue here.
load() gets called from _connect(), which can get called repeatedly from connection monitoring without ever calling disconnect()
This probably means that we need to do the task cleanup either in _on_disconnect(), or avoid re-creating the task in load() if it's already running. The former is probably more correct, as otherwise it will probably spam failure messages during the disconnected state.

bosch_alarm_mode2/panel.py
```diff
@@ -212,36 +306,67 @@ async def _monitor_connection_once(self):
             except asyncio.exceptions.TimeoutError as e:
                 LOG.debug("Connection timed out...")
 
+    async def _login_remote_user(self):
+        creds = int(str(self._passcode).ljust(8, "F"), 16)
```
mag1024, line 310: The int conversion throws if the passcode is not numeric
If the RSC passcode must always numeric, and the passcode we were given isn't that a strong indication that it's actually an automation passcode, and we don't even need to attempt this fallback.

bosch_alarm_mode2/panel.py
```diff
@@ -212,36 +306,67 @@ async def _monitor_connection_once(self):
...
+        bitmask = data[23:].ljust(33, b'\0')
+        # Solution and AMAX panels use one set of arming types, B series panels use another.
+        if data[0] <= 0x24:
+            self._partial_arming_id = AREA_ARMING_STAY1
+            self._all_arming_id = AREA_ARMING_AWAY
+        else:
+            self._partial_arming_id = AREA_ARMING_PERIMETER_DELAY
+            self._all_arming_id = AREA_ARMING_MASTER_DELAY
+        self._supportsSubscriptions = (bitmask[0] & 0x40) != 0
+        self._featureCommandRequestAreaTextCF01 = (bitmask[7] & 0x20) != 0
```
mag1024, line 353: Did we forget to initialize this in the constructor?

bosch_alarm_mode2/panel.py
sanjay900: ah yeah true, did not realise you could use non numeric passcodes in the standard auth

sanjay900:
Ive just sliced the history stuff off into another branch

bosch_alarm_mode2/panel.py
```diff
@@ -212,36 +266,69 @@ async def _monitor_connection_once(self):
...
+            if result[0] == 0x00:
+                if self._passcode.isnumeric():
+                    LOG.info("Authentication failed, trying remote user")
+                    try:
+                        await self._login_remote_user()
+                        LOG.debug("Authentication success!")
+                        return
+                    except Exception:
+                        pass
+                self._connection.close()
```
mag1024, line 289: I think this isn't quite right, because we won't do the close if the AUTHENTICATE returns 0x02.

bosch_alarm_mode2/connection.py
```diff
@@ -19,7 +19,9 @@ def connection_made(self, transport):
 
     def connection_lost(self, exc):
         LOG.info("Connection terminated.")
-        self._on_disconnect()
+        loop = asyncio.get_running_loop()
+        loop.create_task(self._on_disconnect())
```
mag1024, line 23: I think asyncio.create_task is the new preferred way of spawning coros.

But AFAIU, there's no point in spawning a task that we don't await, just to await another task's return.
We should probably keep _on_disconnect() sync, and just do self._poll_task.cancel() without awating.

bosch_alarm_mode2/panel.py
```diff
@@ -308,9 +461,22 @@ def _point_status_consumer(self, data) -> int:
         LOG.debug("Point updated: %s", self.points[point_id])
         return 3
 
+    def _event_summary_consumer(self, data) -> int:
+        priority = data[0]
+        count = _get_int16(data, 1)
+        if count:
+            loop = asyncio.get_running_loop()
+            loop.create_task(self._get_alarms_for_priority(priority))
```
mag1024, line 469: ditto re: using asyncio.create_task

mag1024 (changes requested):
Ok, hopefully the last round of comments!
Thanks for your patience!

bosch_alarm_mode2/connection.py
sanjay900: yeah i wasn't sure if cancelling it like that would cause any problems

bosch_alarm_mode2/panel.py
sanjay900: ah yeah, ill just remove that extra if, i missed that

sanjay900:
Before we merge this, have you got a pypi account so i can transfer that across to you?

mag1024 (approved):
Look great! Thanks for all fixes!
Just created a pypi account under the same username as my github: mag1024

PR #3, opened 2023-05-26
Add support for history

sanjay900:
We use the raw history command as solution and amax panels don't allow using the text events command. 
For panels that do support it, I use the text history command to get human readable text, as it means we don't have to manually parse in those cases.


sanjay900:
@mag1024 Out of curiosity, is this working on your panel now? I fixed that error you encountered the other day

bosch_alarm_mode2/panel.py
```diff
@@ -174,14 +200,14 @@ async def disconnect(self):
                 self._monitor_connection_task = None
         if self._connection: self._connection.close()
 
-    async def area_disarm(self, area_id):
-        await self._area_arm(area_id, AREA_ARMING_DISARM)
+    async def area_disarm(self, area_id, code):
```
mag1024, line 203: Merge error?

bosch_alarm_mode2/const.py
```diff
@@ -142,6 +158,8 @@ class CMD:
     # Alarm memory details
     ALARM_MEMORY_SUMMARY = 0x08
     ALARM_MEMORY_DETAIL = 0x23
+    # History details
+    REQUEST_RAW_HISTORY_EVENTS = 0x15
```
mag1024, line 162: My docs say that 0x15 is deprecated, and 0x63 should be used instead.
There's also, apparently, a version that's supposed to directly return human-readable output (0x16). Have you looked into either of these?

bosch_alarm_mode2/panel.py
```diff
@@ -224,6 +250,25 @@ def _on_disconnect(self):
...
+            # When all events are read, a count of zero is returned
+            # Also, the start is set to the next event id to read
+            if count == 0:
+                self._last_history_msg = start
+                break
+            data = data[5:]
+            for i in range(count):
+                for area in self.areas.values():
+                    area._add_history(self._history_parser(data), self._last_history_msg + i)
+                data = data[self._history_len:]
```
mag1024, line 269: Is see that this is the only use of history len... why do we need to trim it to a different number of events depending on the panel model?

bosch_alarm_mode2/panel.py
```diff
@@ -124,6 +141,10 @@ def __init__(self, host, port, passcode):
         self._connection = None
         self._last_msg = None
         self._poll_task = None
+        self._history_parser = None
+        self._history_len = 0
+        self._last_history_msg = 0
+        self._history = []
```
mag1024, line 147: What do you think of pushing all this state into the History object -- instead of just making a dumb parser?

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,736 @@
+import datetime
+
+
+def _get_int16(data, offset=0):
```
mag1024, line 4: This is a copy of functions already in Panel, right? Let's find a way to share them. Maybe pull them into a util.py?

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,736 @@
...
+    if param2 == 994:
+        user = "PowerUp"
+    if param2 == 995:
+        user = "Telephone"
+    if param2 == 997:
+        user = "Schedule"
+    if param2 == 998:
+        user = "A-Link"
+    if param2 == 999:
+        user = "Installer"
```
mag1024, line 620: Use a dict?

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,736 @@
...
+    "122": "ExInput {param1} Tamper Restore",
+    "123": "PUSH FAIL MOD {param1}",
+    "124": "PUSH RES. MOD {param1}",
+    "125": "Service Mode On",
+    "126": "Service Mode Off",
+
+}
+
+
+def solution_history(date, id, param1, param2):
```
mag1024, line 603: local-only helpers should begin with an _

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,736 @@
...
+    if param2 == 999:
+        user = "Installer"
+
+    return date + SOLUTION_HISTORY_FORMAT[id].format(user=user, param1=param1, param2=param2)
+
+
+def amax_history(date, id, param1, param2):
+    date = f"{date} | "
+    id = str(id)
+    # Amax requires different strings depending on param1 sometimes
```
mag1024, line 628: Oh, wow, this is nasty! Is Amax their oldest panel? Maybe that's why the protocol is so hacked up.

It seems that we're always doing the same action at the end of the day....
I think we can make this simpler by pulling out a helper function decide on a key

def _amax_history_key(id, param1) -> string:
  return id if id in AMAX_HISTORY_FORMAT
  return f"{id}_{param1}" if f"{id}_{param1}" in AMAX_HISTORY_FORMAT
  ...

def amax_history(date, id, param1, param2):
  ...
  key = _amax_history_key(id, param1)
  return date + AMAX_HISTORY_FORMAT[key].format(param1=param1, param2=param2)

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,736 @@
...
+    date = datetime.datetime(year, month, day, hour, minute, second)
+    event_code = _get_int16(data)
+    area = _get_int16(data, 2)
+    param1 = _get_int16(data, 4)
+    param2 = _get_int16(data, 6)
+    param3 = _get_int16(data, 8)
+    return (date, event_code, area, param1, param2, param3)
+
+
+def history_parser(panel):
```
mag1024, line 700: CamelCase, please

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,736 @@
...
+    def _parse(data):
+        if panel <= 0x24:
```
mag1024, line 702: What do you think of creating 3 top separate parsers per model type, and statically creating the right one for the panel we're talking to?

mag1024 (changes requested):

bosch_alarm_mode2/const.py
sanjay900: Yeah so the human readable command isn't supported on the solutions series

bosch_alarm_mode2/panel.py
sanjay900: Different models have the history data in different formats, and we need to know how long the event is in order to chunk it up correctly, though we could possibly do that differently

bosch_alarm_mode2/history.py
sanjay900: Weirdly, the functions in the panel are big endian, while the stuff in history is little and I'm not really sure why that is

bosch_alarm_mode2/const.py
sanjay900: That's the only real reason I had to do this in the first place unfortunately, would have been nice to just use that command for sure!

bosch_alarm_mode2/panel.py
sanjay900: Though this one gets solved automatically once we turn all this into proper classes I suspect

bosch_alarm_mode2/const.py
sanjay900: Ive changed it up to use 0x63 if its available, otherwise it uses 0x16

sanjay900:
Im not sure if there are B or G panels running firmware old enough that they need the history parsing, or if all of them support the human readable events

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,183 @@
...
+    def last_event_id(self):
+        return self._last_event_id
+    
+    def _add_event(self, event, last_event_id):
+        self._events.append(event)
+        self._last_event_id = last_event_id
+    
+    @abc.abstractmethod
+    def _parse_event(self, event):
+        return
```
mag1024, line 36: I guess this structure works.
It might be a bit more clear if we had a single History class, and gave it a model-specific event parser object.
Up to you.

mag1024 (commented):
Haven't gone through everything in history.py yet, but sending out now to start the conversation about how to best expose the history to HA.

bosch_alarm_mode2/panel.py
```diff
@@ -60,6 +63,11 @@ def faults(self): return self._faults
     @property
     def alarms(self): return [ALARM_MEMORY_PRIORITIES[x] for x in self._alarms]
 
+    def _update_history(self, history):
+        self._history = history.events
+        self._last_history_event = history.last_event_id
+        self.history_observer._notify()
+
```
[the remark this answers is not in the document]
mag1024: Alternatively, maybe we could create a Device to represent the panel. I think those can have custom info/actions in a burger sub-menu, and we could have one that shows the log. But that (AFAIK) is not accessible within HA programmatically, so it would be for manual inspection only. I guess that could be ok for history. I can't imagine this big blob of text being particularly useful for automations, etc, anyway.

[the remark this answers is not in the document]
sanjay900: Yeah i don't see why one would build automations based on the history anyways, i think that is mostly just something a person would look at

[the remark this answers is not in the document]
sanjay900: hmm, actually, im not sure if devices let you do that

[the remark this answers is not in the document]
sanjay900: Ive just moved it to its own history sensor

sanjay900:
> The library _seems_ to work (at least it doesn't throw any errors), but with the integration changes I got this:
> 
> > 2023-05-30 00:21:47.694 ERROR (MainThread) [homeassistant.components.alarm_control_panel] Error adding entities for domain alarm_control_panel with platform bosch_alarm
> > Traceback (most recent call last):
> > File "/home/mag/homeassistant/core/homeassistant/helpers/entity_platform.py", line 464, in async_add_entities
> > await asyncio.gather(*tasks)
> > File "/home/mag/homeassistant/core/homeassistant/helpers/entity_platform.py", line 742, in _async_add_entity
> > await entity.add_to_platform_finish()
> > File "/home/mag/homeassistant/core/homeassistant/helpers/entity.py", line 870, in add_to_platform_finish
> > await self.async_added_to_hass()
> > File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/alarm_control_panel.py", line 122, in async_added_to_hass
> > await self._panel.load_history(start_id, history)
> > File "/home/mag/homeassistant/core/config/custom_components/bosch_alarm/bosch_alarm_mode2/panel.py", line 178, in load_history
> > await self._history._load_history(last_event_id, previous_events)
> > TypeError: object NoneType can't be used in 'await' expression

I have fixed this now

bosch_alarm_mode2/history.py
sanjay900: Since its working i don't really want to change it up again

bosch_alarm_mode2/panel.py
```diff
@@ -164,6 +152,22 @@ async def load(self, load_selector):
...
+        self._history._init_history(last_event_id, previous_events)
+        await self._load_history()
+
+    @property
+    def history(self):
+        return self._history.events
+
+    @property
+    def last_history_id(self):
+        return self._history.last_event_id
```
mag1024, line 165: Maybe we can make the history a list of tuples of (event_id, date, description) so that we don't need to keep track of the last_event_id specially?

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,188 @@
...
+        hour = (timestamp >> 12) & 0x1F
+        minute = (timestamp >> 6) & 0x3F
+        second = timestamp & 0x3F
+        date = datetime.datetime(year, month, day, hour, minute, second)
+        event = event[25:].decode()
+        event = f"{date} | {event}"
+        self._add_event(event, BE_INT.get_int16(event, 4))
+        return 25 + text_len
+
+class SolutionAMAXHistory(History):
```
mag1024, line 67: nit: personally I'm a big fan of this: https://www.approxion.com/capital-offenses-how-to-handle-abbreviations-in-camelcase/

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,188 @@
...
+    def _parse_event(self, event):
+        (event_code, date, first_param, second_param) = self._parse_params(event)
+        date = f"{date} | "
+        event_code = str(event_code)
+        user = ""
+        if second_param in SolutionHistory.SOLUTION_USERS:
+            user = SolutionHistory.SOLUTION_USERS[second_param]
+        elif second_param <= 32:
+            user = f"User {second_param}"
+        return date + SOLUTION_HISTORY_FORMAT[event_code].format(user=user, param1=first_param, param2=second_param)
```
mag1024, line 107: nit: line wrap

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,188 @@
...
+class AMAXHistory(SolutionAMAXHistory):
```
mag1024, line 108: line break before all classes, please

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,188 @@
...
+        super().__init__()
+    def _parse_events(self, events):
+        while events:
+            date = events[:11]
+            events = events[11:]
+            time = events[:8]
+            events = events[8:]
+            event = events[:events.index(0)]
+            events = events[len(event):]
+            event = f"{date} {time} | {event}"
```
mag1024, line 178: Oof, the reuse of events, and the similarity of event and events makes this kinda hard to follow.

mag1024:
Long term it's worth studying how the z-wave integration build their devices. They both a have a completely custom statistics popup accessible via the devices' burger menu, and, possibly even more compelling, they emit custom events that get recorded in HA's native Logbook.
image

sanjay900:
> Long term it's worth studying how the z-wave integration build their devices. They both a have a completely custom statistics popup accessible via the devices' burger menu, and, possibly even more compelling, they emit custom events that get recorded in HA's native Logbook. image

well now im tempted to see how the hell they do that because it does look a lot nicer. Getting everything in the log book would also be a nice method to go with as well, i just need to work out how the hell that sorta thing is actually done, and im not 100% sure we can do that with custom timestamps for importing the old events

sanjay900:
> Long term it's worth studying how the z-wave integration build their devices. They both a have a completely custom statistics popup accessible via the devices' burger menu, and, possibly even more compelling, they emit custom events that get recorded in HA's native Logbook. image

Turns out this is something extremely specific to Z-Wave, and isn't just something anybody can do, but what i can do is  set up a device anyways, and then tag the history as a DIAGNOSTIC sensor so that it isnt automatically added to panels or voice assistants

mag1024:
Ah, that's unfortunate. Thanks for investigating!
Anyway, we can fiddle with exactly how we surface the events in HA after we merge the initial version. The main thing I'd like to address before then is to find a way around having to have an init_history call. It seems to me that the current structure is _too_ influenced by the fact that we currently use a separate sensor, and it is the one persisting the history state, and results in a somewhat strange API for the library.
Can we find a way to persist the the history state (last event id) at the integration level, s.t. it can be read at initialization time, and given to the Panel constructor? Alternatively, we could forego state persistence altogether for now -- just always read the entire history on startup (this would also allow us to remove init_history) -- and add load time optimizations later.

sanjay900:
> Ah, that's unfortunate. Thanks for investigating! Anyway, we can fiddle with exactly how we surface the events in HA after we merge the initial version. The main thing I'd like to address before then is to find a way around having to have an init_history call. It seems to me that the current structure is _too_ influenced by the fact that we currently use a separate sensor, and it is the one persisting the history state, and results in a somewhat strange API for the library. Can we find a way to persist the the history state (last event id) at the integration level, s.t. it can be read at initialization time, and given to the Panel constructor? Alternatively, we could forego state persistence altogether for now -- just always read the entire history on startup (this would also allow us to remove init_history) -- and add load time optimizations later.

Ill look into the persistance stuff soon, however i am just going to test what i have with a sol 3000 panel that supports subscriptions, and next week i will be receiving a b series panel as well, so i will test history with both panels and get everything working with the three panels i will have here.

sanjay900:
I have now obtained a b series panel

sanjay900:
> Ah, that's unfortunate. Thanks for investigating! Anyway, we can fiddle with exactly how we surface the events in HA after we merge the initial version. The main thing I'd like to address before then is to find a way around having to have an init_history call. It seems to me that the current structure is _too_ influenced by the fact that we currently use a separate sensor, and it is the one persisting the history state, and results in a somewhat strange API for the library. Can we find a way to persist the the history state (last event id) at the integration level, s.t. it can be read at initialization time, and given to the Panel constructor? Alternatively, we could forego state persistence altogether for now -- just always read the entire history on startup (this would also allow us to remove init_history) -- and add load time optimizations later.

Alright, ive figured out how to use `homeassistant.helpers.storage.Store` and have now decoupled the history persistance from the entity

sanjay900:
Also, now that I'm using developer panels instead of my main house alarm, i ended up setting up a home assistant dev environment. It turns out when you do this, you can easily test both the integration and the module, as you can just clone the integration and symlink it to custom_integrations, and then you can `python -m pip install bosch-alarm-mode2/` and it will install it from the filesystem, and if you then use the "skip pip" option when starting home assistant, it will use the version of the module you just installed, instead of attempting to pull another one.

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,189 @@
...
+            return check
+        check = self._check_history_key(f"{id}_dx2", date, first_param, second_param)
+        if check and first_param <= 108:
+            return check
+        check = self._check_history_key(f"{id}_dx3", date, first_param, second_param)
+        if check and first_param in (150, 151):
+            return check
+        check = self._check_history_key(f"{id}_b4", date, first_param, second_param)
+        if check and first_param in (150, 151):
+            return check
```
mag1024, line 133: probably need a default return here?

bosch_alarm_mode2/panel.py
```diff
@@ -115,32 +101,37 @@ def __repr__(self):
...
         self._supports_command_request_area_text_cf01 = False
         self._supports_command_request_area_text_cf03 = False
+        self._previous_history_events = previous_history_events
 
     LOAD_BASIC_INFO = 1 << 0
     LOAD_ENTITIES = 1 << 1
     LOAD_STATUS = 1 << 2
-    LOAD_ALL = LOAD_BASIC_INFO | LOAD_ENTITIES | LOAD_STATUS
+    LOAD_HISTORY = 1 << 3
+    LOAD_ALL = LOAD_BASIC_INFO | LOAD_ENTITIES | LOAD_STATUS | LOAD_HISTORY
```
mag1024, line 134: I don't think we need a separate load toggle for history.
The 3 different modes are motivated by 3 specific use case:
1) initial connection at integration setup time creates a scratch Panel to verify that the provided connection information is valid, so it uses the minimal "basic info"
2) when creating a new Panel instance normally, we load everything
3) when re-establishing the connection to the panel after an interruption, we need to catch up on state, so we reload the "state"

Under this classification, history is just a sub-case of "state"

bosch_alarm_mode2/panel.py
```diff
@@ -153,6 +144,10 @@ async def load(self, load_selector):
         if load_selector & self.LOAD_ENTITIES:
             await self._load_areas()
             await self._load_points()
+        if load_selector & self.LOAD_HISTORY:
+            self._history._init_history(self._previous_history_events)
+            if self._supports_subscriptions:
+                await self._load_history()
```
mag1024, line 150: let's do the init in the in the constructor, without storing previous events in a member
and is there a reason to make the _load_history() conditional?

bosch_alarm_mode2/history.py
```diff
@@ -0,0 +1,189 @@
...
+    
+    def _add_event(self, event, last_event_id):
+        self._events.append((last_event_id, event))
+        self._panel.history_observer._notify()
+    
+    @abc.abstractmethod
+    def _parse_event(self, event):
+        return
+    
+    def _init_history(self, events):
```
mag1024, line 30: This is called by the Panel, so it should be "public", right? (i.e. remove "_")

bosch_alarm_mode2/panel.py
```diff
@@ -164,6 +159,10 @@ async def load(self, load_selector):
                 LOG.info(
                     "Panel does not support subscriptions, falling back to polling")
 
+    @property
+    def history(self):
```
mag1024, line 163: a return type annotation might be useful here to avoid forcing users of Panel to dig into the implementation of the history class

bosch_alarm_mode2/panel.py
```diff
@@ -309,8 +324,16 @@ async def _basicinfo(self):
...
         self._supports_command_request_area_text_cf01 = (bitmask[7] & 0x20) != 0
         self._supports_command_request_area_text_cf03 = (bitmask[7] & 0x08) != 0
-        # Check if serial read command is supported before sending it
-        if (bitmask[13] & 0x04) != 0:
+        supports_command_request_serial_read = (bitmask[13] & 0x04) != 0
+        supports_command_request_history_text = (bitmask[5] & 0x80) != 0
+        supports_command_request_history_raw_ext = (bitmask[16] & 0x02) != 0
+        if not supports_command_request_history_text:
+            self._history = construct_raw_parser(data[0], self)
+            if supports_command_request_history_raw_ext:
```
mag1024, line 332: self._history_cmd = ... if ... else ... 
might be more idiomatic

bosch_alarm_mode2/panel.py
```diff
@@ -436,44 +459,49 @@ async def _subscribe(self):
...
             asyncio.create_task(self._get_alarms_for_priority(priority))
         else:
             # Alarms are no longer triggered, clear
             for area in self.areas.values():
                 area._set_alarm(priority, False)
         return 3
+    
+
+    def _event_history_consumer(self, data) -> int:
+        return self._history._parse_subscription_event(data)
```
mag1024, line 498: This method (correctly) doesn't have the leading _

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "bosch-alarm-mode2"
-version = "0.3.20"
+version = "0.0.0.dev3"
```
mag1024, line 3: Remember to revert please

mag1024 (commented):
Overall looks good, thank you! Only minor comments this time.

mag1024:
> I have now obtained a b series panel

Cool! It's very confusing that my panel **B**8512G is part of the G-series. Who came up with this naming??
BTW, is there any chance your Bosch contacts can hook me up with whatever's necessary to run the configuration software for my panel? I've tried to get this to work a while back, and wasn't able to, because I think it requires some kind of a hardware security dongle.

mag1024:
> Since you set the device up, i also specified that the history sensor is a diagnostic sensor, and so it gets put into the diagnostic section of the device instead of being listed with the other sensors.

Oh nice. That seems good for now. I randomly met Paulus (the HA founder) the other day, and he mentioned something about how integrations can even have custom dashboards now. I feel like there are a lot of advanced non-documented options for what one can do, so maybe eventually we'll find something better.

bosch_alarm_mode2/panel.py
sanjay900: that doesn't work as the history object will change once we know what type of history the panel supports. I suppose this is where your idea of splitting out parsing from the history object would probably help

pyproject.toml
sanjay900: Ill revert now, i don't actually need to do this sort of thing anymore

sanjay900:
> > I have now obtained a b series panel
> 
> Cool! It's very confusing that my panel **B**8512G is part of the G-series. Who came up with this naming?? BTW, is there any chance your Bosch contacts can hook me up with whatever's necessary to run the configuration software for my panel? I've tried to get this to work a while back, and wasn't able to, because I think it requires some kind of a hardware security dongle.

what did it do when you tried to use it? I notice the the guide for the B series panel also mentions the same key, but it isnt necessary, I did however find that when i was configuring it with RPS, it refused to let me do the initial setup (giving me a red X) unless i put in my public IP address, which i then was able to change back to my local one later for testing. You also should be able to just use a usb A to A cable for programming directly as well from what i gather.

sanjay900:
> > I have now obtained a b series panel
> 
> Cool! It's very confusing that my panel **B**8512G is part of the G-series. Who came up with this naming?? BTW, is there any chance your Bosch contacts can hook me up with whatever's necessary to run the configuration software for my panel? I've tried to get this to work a while back, and wasn't able to, because I think it requires some kind of a hardware security dongle.

Haha that naming confused the hell out of me too when i was implementing that history stuff, but that reminds me to add the fact that we now know the B5512 works

sanjay900:
I have now also been able to confirm that subscription based history events are parsing correctly as well, and i have simplified the parsing of text based events

mag1024:
The parser changes look like they worked out great -- thank you!

mag1024:
> I did however find that when i was configuring it with RPS, it refused to let me do the initial setup (giving me a red X) unless i put in my public IP address, which i then was able to change back to my local one later for testing.

Ah, that could very well be it! I'll have to try that the next time I get around to playing with it.

mag1024:
The history loading is so slow on my panel, that it trips hearbeat and resets the connection, causing it to restart the load 🤦

mag1024:
Maybe we should limit it to something reasonable like 1 month of events of backlog -- otherwise it's loading stuff from 2021, which doesn't seem particularly useful.....

sanjay900:
Annoyingly, all we can do is specify a event id to start from, there's no way to specify a date specifically.

mag1024:
Yeah, I'm in strongly in favor of not adding any parameters that we don't absolutely need, so hardcoding a sane default for the number of events to load sounds pretty reasonable.
In the meantime, I experimented with forcing the use of the binary history format on my panel to see if it's any faster, and found that it totally doesn't work. There are so many errors, that I suspect that there's something totally off with the parsing: either wrong endian-ness, or maybe something off with the byte offsets.
Dropping some sample errors here in case you get around to looking into it before I do: [history-errors.txt](https://github.com/mag1024/bosch-alarm-mode2/files/11780438/history-errors.txt)

mag1024:
Oh, yeah, it's totally the endian-ness: switching to BE fixes it. Can you please check if your B-series uses LE or BE in binary mode?

mag1024:
BTW, have you looked into Protocol 4 at all? Seems like it basically uses 16 bit lengths, instead of 8 bit, allowing it to return much larger responses -- up to 23 per history response. I tried it out by hacking some stuff in connection.py, and the panel seems to respond with... something. But it doesn't match what we're currently expecting

sanjay900:
I didn't really look at the other formats tbh, but tbh I kind of like the idea of just limiting it to 30 responses or so.
I implemented that logic here: https://github.com/mag1024/bosch-alarm-mode2/pull/7

PR #4, opened 2023-05-30
Fix check for serial number command support

sanjay900:
While looking at the history support, i realised that i had got the index slightly wrong.


mag1024:
Thanks! I'm adding you as a collaborator for both projects.
Feel free to submit bug fixes like this directly, but please continue to use branches + reviews for feature development.

PR #6, opened 2023-06-16
Only show sensors and areas that are actually enabled

sanjay900:
`_load_authorised_entities` will only return enabled entities. We already had this logic on cf01, so apply the same logic to cf03


sanjay900:
Is this the sort of thing you'd prefer as a MR, or just to be committed straight in?

mag1024:
Yeah, anything that's a localized fix like this, please feel free to submit it directly.

PR #7, opened 2023-06-17
Limit the amount of history pulled back

sanjay900:
Instead of pulling the entire panels history back, only load the last 30 events.
Also, update the `_last_msg` when retrieving history, to make sure that loading lots of events at once won't ever cause issues.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -236,15 +236,14 @@ class Panel:
         return start, count
 
     async def _load_history(self):
-        if self._history.last_event_id == NO_EVENTS:
-            start, count = await self._read_history(self._history.last_event_id)
-            start -= EVENT_LOOKBACK_COUNT
-            start = max(0, start)
-            start, count = await self._read_history(start)
-            if count == 0:
-                return
+        load_old_events = self._history.last_event_id == NO_EVENTS
         while True:
             start, count = await self._read_history(self._history.last_event_id)
+            if load_old_events:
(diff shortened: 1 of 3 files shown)


sanjay900:
@mag1024 Does 30 events sound like a good number, or do you think we would be better with more or less than that?

bosch_alarm_mode2/panel.py
```diff
@@ -219,18 +220,30 @@ def _on_disconnect(self):
...
-            # When all events are read, a count of zero is returned
-            # Also, the start is set to the next event id to read
-            if count:
-                data = data[5:]
-            self._history.parse_polled_events(start, data, count)
+            start, count = await self._read_history(self._history.last_event_id)
+            if load_old_events:
+                start -= EVENT_LOOKBACK_COUNT
+                start = max(0, start)
+                start, count = await self._read_history(start)
```
mag1024, line 245: Oh heh, the multiple read_history calls in this function make it quite tricky to figure out what's going on.

I think we can push all this inside the history object:
1) if events are empty, return 0xFF as the last_event_id
2) if history gets an empty data event, and there are no events, create a sentinel event at _id - history_read_limit_
3) push parsing of count/start into _parse_polled_events_, and have it return a bool to indicate if the load loop should continue -- this way we can avoid exiting the loop after 2

mag1024 (changes requested):

sanjay900:
Also can confirm, the endinaness fix works on my b series panel as well

PR #8, opened 2023-06-19
History limiting v2

mag1024:
Builds on #7, but includes the refinement of avoiding an explicit sentinel event by returning the next id to use from _parse_polled_events_. A couple of other misc cleanups.


sanjay900:
Yeah this looks much better, I'm happy for this to be merged, I'm not near my PC or panel for the next week's so I can't test it but it looks good

PR #9, opened 2023-06-20
Protocol4

mag1024:
Experimental protocol 4 support.
Seems to work, and returns up to 23 history events per request.
No idea how broadly supported it is though.
@sanjay900, when you're back, could you please check if the panels you have access to support this.
No rush, and thanks in advance :))

### DIFF
--- a/bosch_alarm_mode2/connection.py
+++ b/bosch_alarm_mode2/connection.py
@@ -1,13 +1,14 @@
 import asyncio
 import logging
 
-from .const import *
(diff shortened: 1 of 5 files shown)


sanjay900:
Should be able to query for support using whatareyou and support this sort of thing conditionally?

sanjay900:
Seems to work on all my panels. The solution ones stay on protocol 1, while the B series panel does negotiate and use protocol 4.

PR #10, opened 2023-06-20
History cleanups

mag1024:
Some work-in-progress attempts at cleanups
* Attempts to make the Amax history parsing logic less redundant -- this is quite tricky for me to test, so not totally sure if it works... 
* Breaks date out into a separate field in the history event

@sanjay900, PTAL when you're back.


sanjay900:
I wouldn't worry too much about AMAX, we can test that once i get my hands on a panel (and hopefully i will be getting my hands on one soon), and i suspect there is a bunch more that needs to go into AMAX support anyways from what i understand

Issue #11, opened 2023-06-22
Consider dropping support for text history


mag1024:
It seems to be slower (i.e. fewer events returned by response), and harder to understand (more terse, using abbreviations).
Example:

text:
```
[5005] 2023-06-16 20:48:00 | Network Cable Open On-board Ethernet
[5006] 2023-06-16 20:48:00 | Network Cable Good On-board Ethernet
[5007] 2023-06-17 20:25:00 | User Code Tamper Ar-1 P240,Us-5002
[5008] 2023-06-18 02:02:00 | Sked Report Exec Sk-3
[5009] 2023-06-18 02:02:00 | Test Report
```

raw:
```
[5005] 2023-06-16 20:48:00 | Network Cable Disconnected, SDI 2 Device: 399
[5006] 2023-06-16 20:48:00 | Network Cable Connected, SDI 2 Device: 399
[5007] 2023-06-17 20:25:00 | User passcode tamper - too many attempts, Area: 0, User ID: 5002
[5008] 2023-06-18 02:02:00 | Test Report Sked has executed., Sked: 2
[5009] 2023-06-18 02:02:00 | Test Report - System off-normal, non-expanded status
```

Seems like it might be a negative value add.

sanjay900:
Seems reasonable, tbh I had that thought when I was implementing it too

sanjay900:
Do note that if we do this, we will also want to change subscription based history events to parsing the raw history data as well.

sanjay900:
Well, i now have a solution panel that sends subscription events, and i will have to change the subscription code a little as the solution history events do not contain a timestamp in the text section

sanjay900:
Interesting... History reading NAKs if the panel is ARMED

mag1024:
Weird -- I wonder if that's a bug or a feature.
We should probably catch the naks around the history poll to avoid rapid-fire log spam, right?

BTW, I was thinking we should really introduce an exception type for naks, and catch it explicitly instead of relying on the catch all.

PR #12, opened 2023-06-25
Implement raw version of subscription history

sanjay900:
While thinking about #11, i realised that we would want a raw version of the subscription history events as well, so I have split the history parser up a little to allow reusing that logic for subscriptions as well.
This also means things won't crash if something goes wrong while parsing subscription history events.

### DIFF
--- a/bosch_alarm_mode2/history.py
+++ b/bosch_alarm_mode2/history.py
@@ -50,6 +50,11 @@ class History:
             self._parser = AmaxHistory()
         else:
             self._parser = BGHistory()
+    
+    def _append_error(self, id, excp):
+        error_str = f"parse error: {repr(excp)}"
+        LOG.error("History event " + error_str)
+        self._events.append(HistoryEvent(id, datetime.now(), error_str))
 
     def parse_polled_events(self, event_data):
         count = event_data[0]
@@ -67,9 +72,7 @@ class History:
                 LOG.debug(e)
             self._events.extend(events)
         except Exception as excp:
-            error_str = f"parse error: {repr(excp)}"
-            LOG.error("History event " + error_str)
-            self._events.append(HistoryEvent(start + count, datetime.now(), error_str))
+            self._append_error(start + count, excp)
         return self.last_event_id
 
     def parse_subscription_event(self, raw_event):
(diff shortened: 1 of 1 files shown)


sanjay900:
I was able to validate this on my b series panel by initiating a panel sensor reset, and seeing that event come through

bosch_alarm_mode2/history.py
```diff
@@ -19,6 +19,14 @@ class HistoryEvent(NamedTuple):
     def __repr__(self):
         return f"[{self.id}] {self.date} | {self.message}"
 
+class HistoryEventParams(NamedTuple):
+    date: datetime
+    event_code: int
```
mag1024, line 24: We actually seem to treat this as str in most places

bosch_alarm_mode2/history.py
```diff
@@ -66,12 +74,16 @@ def parse_polled_events(self, event_data):
 
     def parse_subscription_event(self, raw_event):
         text_len = BE_INT.int16(raw_event, 23)
+        event_id = BE_INT.int32(raw_event)
```
mag1024, line 77: should probably length check raw_event -- or put the try around the whole thing

bosch_alarm_mode2/history.py
```diff
@@ -66,12 +74,16 @@ def parse_polled_events(self, event_data):
...
-        date = datetime.strptime(f"{date} {time}", "%m/%d/%Y %I:%M%p")
-        self._events.append((BE_INT.int16(raw_event, 4), date, message))
+        try:
+            e = self._parser.parse_subscription_event(raw_event)
+            LOG.debug(e)
+            self._events.append(e)
+        except Exception as excp:
+            error_str = f"parse error: {repr(excp)}"
+            LOG.error("History event " + error_str)
+            self._events.append(HistoryEvent(event_id + 1, datetime.now(), error_str))
```
mag1024, line 86: Share this with parse_polled_events?

mag1024 (approved):
Yeah, good idea!
Let's get drop the TextHistory next, and we can probably remove the abstractmethods as well.

sanjay900:
Yeah, i like the idea of having a commit around that has all the text stuff implemented, before we drop it just in case there is ever a point we need it all working again for some reason

mag1024:
Thanks!
Next time we touch this file, we should wrap some of the extra long lines -- pylint should point them out.

PR #13, opened 2023-06-27
Remove text history, fix up subscription history timestamp parsing

sanjay900:
The subscription events return a timestamp in an internal format it seems, which does change a bit between the different panels. 
Sadly, the solution panels don't even give you a date as part of the text format either, so we can't even rely on parsing that either.
I don't think amax panels even support subscriptions, nor do i expect there to ever be a dev firmware that adds that functionality, but it'll likely be the same format as solution either way if they do add one, so i've just made that assumption here

### DIFF
--- a/bosch_alarm_mode2/history.py
+++ b/bosch_alarm_mode2/history.py
@@ -73,7 +73,7 @@ class History:
         event_length = len(event_data) // count
         for i in range(start, start + count):
             try:
-                e = self._parser.parse_raw_event(i, event_data)
+                e = self._parser.parse_polled_event(i, event_data)
                 LOG.debug(e)
                 self._events.append(e)
                 event_data = event_data[event_length:]
@@ -115,7 +115,7 @@ class HistoryParser:
             date=date, event_code=event_code, area=area, param1=param1, param2=param2, param3=param3)
         return HistoryEvent(BE_INT.int32(raw_event) + 1, *self._parse_event(params))
 
-    def parse_raw_event(self, id, event_data):
+    def parse_polled_event(self, id, event_data):
         return HistoryEvent(id, *self._parse_event(self._parse_event_params(event_data)))
 
     @abc.abstractmethod
--- a/bosch_alarm_mode2/history.py
+++ b/bosch_alarm_mode2/history.py
@@ -94,15 +94,6 @@ class History:
 class HistoryParser:
     __metaclass__ = abc.ABCMeta
 
-    @abc.abstractmethod
-    def parse_events(self, start, event_data, count) -> [HistoryEvent]:
-        pass
-    
-    @abc.abstractmethod
-    def parse_subscription_event(self, raw_event) -> HistoryEvent:
-        pass
-
-class RawHistory(HistoryParser):
     def parse_events(self, start, event_data, count) -> [HistoryEvent]:
         events = []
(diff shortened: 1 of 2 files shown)


bosch_alarm_mode2/history.py
```diff
@@ -137,21 +115,24 @@ def parse_subscription_event(self, raw_event) -> HistoryEvent:
...
         param3 = BE_INT.int16(raw_event, 12)
-        text_len = BE_INT.int16(raw_event, 23)
-        total_len = 25 + text_len
-        line = raw_event[25:total_len - 1].decode()
-        date, time, _ = re.split(r"\s+", line, 2)
-        date = datetime.strptime(f"{date} {time}", "%m/%d/%Y %I:%M%p")
-        params = HistoryEventParams(date=date, event_code=event_code, area=area, param1=param1, param2=param2, param3=param3)
+        timestamp = BE_INT.int32(raw_event, 14)
+        params = HistoryEventParams(date=self._parse_subscription_event_timestamp(
+            timestamp), event_code=event_code, area=area, param1=param1, param2=param2, param3=param3)
```
mag1024, line 120: This line wrapping looks really funky.... a temporary for date might help

bosch_alarm_mode2/history.py
```diff
@@ -91,43 +95,17 @@ def parse_subscription_event(self, raw_event):
...
-        # Not sure why, but there is an extra 0 in subscription text
-        date, time, _, message = re.split(r"\s+", line, 3)
-        date = datetime.strptime(f"{date} {time}", "%m/%d/%Y %I:%M%p")
-        return HistoryEvent(BE_INT.int32(raw_event) + 1, date, message)
-
-class RawHistory(HistoryParser):
     def parse_events(self, start, event_data, count) -> [HistoryEvent]:
         events = []
         event_length = len(event_data) // count
         for i in range(count):
```
mag1024, line 106: Now that there's a single way to decide the split the event_data into individual events events, let's pull the loop into History.parse_polled_events. This way we can try/except around the parsing of individual events, and a single bad event won't cause the whole batch to be discarded, as it does currently.

bosch_alarm_mode2/history.py
```diff
@@ -51,7 +54,7 @@ def init_raw_history(self, panel_type):
             self._parser = AmaxHistory()
```
mag1024, line 54: The "raw" in init_raw_history no longer makes sense now that everything is raw -- may something like init_for_panel?

bosch_alarm_mode2/history.py
```diff
@@ -168,19 +149,38 @@ def _parse_sol_amax_params(event):
...
+    minute = timestamp & 0x3F
+    hour = (timestamp >> 6) & 0x1F
+    day = (timestamp >> 11) & 0x1F
+    month = (timestamp >> 16) & 0x0F
+    year = 2000 + ((timestamp >> 20) & 0x1F)
+    second = (timestamp >> 26)
+    return datetime(year, month, day, hour, minute, second)
+
+
+class SolutionHistory(HistoryParser):
```
mag1024, line 163: These classes should all probably be named FooHistory*Parser*

mag1024 (commented):
A really nice simplification!

mag1024:
Thank you!

PR #14, opened 2023-07-05
Add check for using an automation passcode on a solution / amax panel

sanjay900:
Not quite sure what error message to use here - and we might even want  to roll this into the idea of using proper exceptions too. This however does work, on my solution panel this catches me trying to use an automation passcode instead of the RSC+ passcode.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -228,7 +228,7 @@ class Panel:
 
     async def _load_history(self):
         # Don't retrieve history when armed, as panels do not support this.
-        if any(area.is_armed() for area in self.areas):
+        if any(area.is_armed() for area in self.areas.values()):
             return
         start_size = len(self.events)
         start_t = time.perf_counter()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "bosch-alarm-mode2"
-version = "0.3.22"
+version = "0.3.24"
 description = ""
 authors = ["Michael Grigoriev <mag@luminal.org>"]
 


bosch_alarm_mode2/panel.py
```diff
@@ -322,6 +322,11 @@ async def _basicinfo(self):
         if data[0] <= 0x24:
             self._partial_arming_id = AREA_ARMING_STAY1
             self._all_arming_id = AREA_ARMING_AWAY
+            # This command fails when used with the wrong type of passcode
+            try:
```
mag1024, line 326: Hm, doesn't the the LOGIN_REMOTE_USER return something in the response to let you know that the password is incorrect?

If we do need this, shouldn't this go somewhere with authentication, like _login_remote_user, rather than basicinfo?

bosch_alarm_mode2/panel.py
```diff
@@ -322,6 +322,11 @@ async def _basicinfo(self):
...
+                await self._connection.send_command(CMD.REQUEST_CONFIGURED_AREAS)
+            except:
+                raise Exception("Incorrect code used - Solution and AMAX series panels require the RSC+ passcode, not the automation passcode.")
```
mag1024, line 329: Should probably be PermissionError

bosch_alarm_mode2/panel.py
sanjay900: The panel does actually authenticate successfully if you use the automation passcode on a solution panel, but then commands don't work correctly, its quite odd.

sanjay900: It would have to go in `authenticate` as we would never hit the authentication failed error condition if someone happens to use a valid automation passcode on a solution panel.

mag1024:
Hm, I see.... Weird! What do the subsequent commands return for the error code? Is it something sensible like a NACK with a "No Authority" or "Invalid Remote ID" error?

sanjay900:
> Hm, I see.... Weird! What do the subsequent commands return for the error code? Is it something sensible like a NACK with a "No Authority" or "Invalid Remote ID" error?

Unfortunately not, it just gives us `Exception: ('NACK: ', 'Execution Function Sequence Wrong')`

mag1024:
As much as I liked the approach of automatically trying both auth methods, s.t. we don't have to maintain a list of which panels support which auths, at this point, maybe it's time to go back to doing basicinfo before authentication, and choosing the authentication command based on the panel type. Otherwise it's getting weird to the point that the cure is starting to seem worse than the disease.

sanjay900:
Seems reasonable, i was hoping by now i would have a amax panel so i could verify what authentication is required for it, but that has not arrived yet unfortunately.

PR #15, opened 2023-09-03
Don't update panel history if the panel is armed

sanjay900:
Better fix for https://github.com/mag1024/bosch-alarm-homeassistant/issues/10

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -467,6 +467,8 @@ class Panel:
         area_id = BE_INT.int16(data)
         area_status = self.areas[area_id].status = data[2]
         LOG.debug("Area %d: %s" % (area_id, AREA_STATUS.TEXT[area_status]))
+        if area_status==AREA_STATUS.DISARMED:
+            asyncio.create_task(self._load_history())
         return 3
 
     def _area_ready_consumer(self, data) -> int:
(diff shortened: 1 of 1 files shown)


sanjay900:
just catching the exception means we will flood users logs, while it would be better to just not support retriving history  when the panel is not armed.

bosch_alarm_mode2/panel.py
```diff
@@ -465,6 +467,8 @@ def _area_on_off_consumer(self, data) -> int:
         area_id = BE_INT.int16(data)
         area_status = self.areas[area_id].status = data[2]
         LOG.debug("Area %d: %s" % (area_id, AREA_STATUS.TEXT[area_status]))
+        if area_status==AREA_STATUS.DISARMED:
```
mag1024, line 470: Spaces around ==, please. Alternatively, we could not check at all -- we're already relying on _load_history to check if _all_ areas are now disarmed.

mag1024, line 470: Please add a comment that this is necessary because the initial history load may have been skipped if some areas were armed during initialization.

mag1024 (changes requested):

mag1024:
Thank you!

sanjay900:
@mag1024 did you end up releasing another version of the home assistant integration with these changes?

mag1024:
Nope, I have not. I can try to do it later in the week -- or if you get to it sooner, please feel free.

PR #17, opened 2023-09-22
Change how we handle passcodes for the various panels

sanjay900:
Continuation of https://github.com/mag1024/bosch-alarm-mode2/pull/14, just needed to rebase so i wasn't making these changes from my master branch


mag1024:
So like I mentioned on #14, I think we should probably restructure the way we do authentication: go back to the idea of doing basicinfo before authentication, and choosing the authentication command based on the panel type.

sanjay900:
Yeah 100%, was just leaving this open so I remember to come back and do that unless you get to it first, just needed those changes not on my master haha

sanjay900:
Unfortunately, if the passcode is wrong, we do just get a `Exception: ('NACK: ', 'Non-specific error')`, but its what it is

sanjay900:
Since i was updating stuff related to panel information, i also figured i would take this chance to get the proper firmware version instead of just the protocol version as well.

bosch_alarm_mode2/panel.py
```diff
@@ -117,16 +117,19 @@ def __init__(self, host, port, passcode):
...
         self._history_cmd = None
         self.areas = {}
         self.points = {}
         self._partial_arming_id = AREA_ARMING_PERIMETER_DELAY
         self._all_arming_id = AREA_ARMING_MASTER_DELAY
+        self._supports_serial = False
         self._supports_subscriptions = False
         self._supports_command_request_area_text_cf01 = False
         self._supports_command_request_area_text_cf03 = False
+        self._is_solution_or_amax = False
```
mag1024, line 132: It would more consistent with other panel-specific options to name this something like _supports_automation_user  (or the opposite -- _supports_remote_user_login)

bosch_alarm_mode2/panel.py
```diff
@@ -281,33 +286,37 @@ async def _monitor_connection_once(self):
...
-                pass
-
         self._connection.close()
         error = ["Not Authorized", "Authorized",
                 "Max Connections"][result[0] if result else 0]
         raise PermissionError("Authentication failed: " + error)
 
+    async def _authenticate(self):
+        if self._is_solution_or_amax:
+            if not self._passcode.isnumeric():
```
mag1024, line 312: Let's push this error checking inside _authenticate_remote_user -- and maybe line wrap these long error strings.

bosch_alarm_mode2/panel.py
```diff
@@ -322,24 +331,34 @@ async def _basicinfo(self):
...
+        if self._supports_serial:  # supports serial read
             data = await self._connection.send_command(
                 CMD.PRODUCT_SERIAL, b'\x00\x00')
             self.serial_number = int.from_bytes(data[0:6], 'big')
+        if self._supports_status:
+            data = await self._connection.send_command(
+                CMD.REQUEST_PANEL_SYSTEM_STATUS)
+            version = data[0]
+            revision = int.from_bytes(data[1:2], 'big')
+            self.firmware_version = f"v{version}.{revision}"
```
mag1024, line 361: nit: let's keep this consistent with how we format the protocol version:

'v%d.%d' % (..., ...)

mag1024 (commented):
I'm pleasantly surprised at how neat this turned out!

PR #18, opened 2023-10-01
Output support

sanjay900:
I've added support for outputs. If you configure an output as a "Remote output" on a solution 3000, then it can actually be configured and controlled by mode2.

This seems to work on both my solution 3000 panel and my B5512. 

It seems that the solution 3000 subscription status gives us a raw output number, as the panel has other outputs besides remote ones, and the first remote output would end up being output 6, hence the need to handle outputs differently there.


sanjay900:
Implementing this was a request from Bosch. 
Both the async messages and the polling seem to be working.

sanjay900:
Oh yeah FYI i did learn one thing the other week you will probably find useful
when testing mode2 stuff, you can just clone it to your home assistant config directory, and it will use that version of the python module.

sanjay900:
```
Model: B5512 (US1B)
Protocol version: v5.234
Serial number: 153966761042037
Areas:
{1: Area 1: Disarmed [Not Ready] (0)}
Points:
{1: Point 1 : Normal}
Outputs:
{3: Output C (3): Inactive, 2: Output B (2): Inactive, 1: Output A (1): Inactive}
```

bosch_alarm_mode2/panel.py
```diff
@@ -439,6 +480,30 @@ async def _load_entity_status(self, status_cmd, entities):
...
+                id = index + (8 - i)
+                if id in self.outputs:
+                    self.outputs[id].status = int(b & 1 != 0)
+                b >>= 1
+            index += 8
+        
+        # The OUTPUT_STATUS command only returns data for the highest active output.
+        # This means that we need to fill in any other outputs.
+        for id, output in self.outputs.items():
+            if id > index * 8:
```
mag1024, line 498: Is this accurate?
If the highest output is 9, don't we need to clear 10-15?

How about we just reset all the output values before we process the response? That seems harder to get wrong.

bosch_alarm_mode2/panel.py
```diff
@@ -323,6 +354,9 @@ async def _basicinfo(self):
         if data[0] <= 0x24:
             self._partial_arming_id = AREA_ARMING_STAY1
             self._all_arming_id = AREA_ARMING_AWAY
+            # It seems the async commands give us all outputs, while every other command
```
mag1024, line 357: When you say "every other command" -- does that also include the name getting, or just the output setting?

bosch_alarm_mode2/panel.py
```diff
@@ -439,6 +480,30 @@ async def _load_entity_status(self, status_cmd, entities):
...
+        # The OUTPUT_STATUS command only returns data for the highest active output.
+        # This means that we need to fill in any other outputs.
+        for id, output in self.outputs.items():
+            if id > index * 8:
+                output.status = OUTPUT_STATUS.INACTIVE 
+
+
+    async def _set_output_state(self, output_id, state):
+        request = bytearray([output_id, state])
+        self.outputs[output_id].status = state
```
mag1024, line 504: It seems much more robust to rely to rely on the subscription/poll to pick up the change, rather than applying the status eagerly

bosch_alarm_mode2/panel.py
```diff
@@ -456,7 +521,7 @@ async def _subscribe(self):
         data += IGNORE    # config change
         data += SUBSCRIBE # area on/off
         data += SUBSCRIBE # area ready
-        data += IGNORE    # output status
+        data += SUBSCRIBE    # output status
```
mag1024, line 524: nit: fix the comment alignment please

bosch_alarm_mode2/panel.py
```diff
@@ -483,6 +548,14 @@ def _area_ready_consumer(self, data) -> int:
             area_id, AREA_READY[ready_status], faults))
         return 5
 
+    def _output_status_consumer(self, data) -> int:
+        output_id = BE_INT.int16(data) - self._async_output_offset
+        if output_id not in self.outputs:
+            return 3
```
mag1024, line 554: Normally I'm a fan of early returns as they reduce indentation, but in this case, since we have to return a value, I would write
```
if output_id in self.output:
  ...
  ...
return 3
```

bosch_alarm_mode2/panel.py
```diff
@@ -99,6 +99,23 @@ def reset(self):
...
+    
+
+class Output(PanelEntity):
+    def __init__(self, name = None, status = OUTPUT_STATUS.UNKNOWN):
+        PanelEntity.__init__(self, name, status)
+
+    def is_active(self) -> bool:
+        return self.status == OUTPUT_STATUS.ACTIVE
+    
+    def is_inactive(self) -> bool:
```
mag1024, line 111: Seems like we never use this

bosch_alarm_mode2/panel.py
```diff
@@ -123,11 +140,13 @@ def __init__(self, host, port, passcode):
         self._history_cmd = None
         self.areas = {}
         self.points = {}
+        self.outputs = {}
         self._partial_arming_id = AREA_ARMING_PERIMETER_DELAY
         self._all_arming_id = AREA_ARMING_MASTER_DELAY
         self._supports_subscriptions = False
         self._supports_command_request_area_text_cf01 = False
         self._supports_command_request_area_text_cf03 = False
+        self._async_output_offset = 0
```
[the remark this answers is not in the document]
sanjay900: maybe _output_subscription_start_index, subscription was the word i was after haha

bosch_alarm_mode2/panel.py
sanjay900: all the other output related commands including name getting only give us information on the remote outputs for the solution boards, while the subscription commands give us the status of all outputs. Its a bit bizarre but its how it is haha

bosch_alarm_mode2/panel.py
sanjay900: Yeah, the main problem is that there is a delay there, which left the home assistant interface looking a bit weird as i would see  switches jump between states

bosch_alarm_mode2/panel.py
sanjay900: is_active is being used by the home assistant integration at the moment, it just seemed cleaner than doing the check in the integration. is_inactive isn't being used though, so we can remove that

bosch_alarm_mode2/panel.py
sanjay900: i also found if i turned several outputs on and off a bunch of times, ocassionally i would just not see a subscription event for some of the outputs, but they would actually be toggled on the panel

sanjay900: Ah, if i use a semaphore on set_output_state, that problem seems to go away completely.

sanjay900: There is still a weird delay with receiving events, but that does actually appear to only be an issue for the solution panels, so ill check with bosch on that, but that isnt something we need to do anything special about here.

mag1024:
> when testing mode2 stuff, you can just clone it to your home assistant config directory, and it will use that version of the python module.

So just `config/bosch-alarm-mode2`? Thanks, I'll have to try that!

bosch_alarm_mode2/panel.py
```diff
@@ -123,11 +137,14 @@ def __init__(self, host, port, passcode):
...
         self.areas = {}
         self.points = {}
+        self.outputs = {}
         self._partial_arming_id = AREA_ARMING_PERIMETER_DELAY
         self._all_arming_id = AREA_ARMING_MASTER_DELAY
         self._supports_subscriptions = False
         self._supports_command_request_area_text_cf01 = False
         self._supports_command_request_area_text_cf03 = False
+        self._output_subscription_start_index = 0
+        self._output_semaphore = asyncio.Semaphore(1)
```
mag1024, line 147: Do we still need this now that we rely on subscriptions/polling to actually update our internal output state? I can't think of any situations that would cause a race.

bosch_alarm_mode2/panel.py
```diff
@@ -439,6 +477,16 @@ async def _load_entity_status(self, status_cmd, entities):
             entities[BE_INT.int16(response)].status = response[2]
             response = response[3:]
 
+    async def _load_output_status(self):
+        enabled = await self._load_enabled_entities(CMD.OUTPUT_STATUS, "OUTPUT")
```
mag1024, line 481: The reuse looks like it worked out nicely, but now it's very confusing that we're using a function called "enabled" to fetch status.
How about we rename it to something like _load_entity_set -- and also change it so that it doesn't do anything with names. The fallback name generation if can happen directly at the end of _load_names.

bosch_alarm_mode2/panel.py
```diff
@@ -323,6 +352,11 @@ async def _basicinfo(self):
         if data[0] <= 0x24:
             self._partial_arming_id = AREA_ARMING_STAY1
             self._all_arming_id = AREA_ARMING_AWAY
+            # The solution panels only offer control over remote outputs. 
+            # For this reason, all other output types don't show up in commands at all.
```
mag1024, line 356: Maybe something like

```
# The solution panels only offer control over remote outputs.
# For most commands, output 0 is the first remote output.
# However, subscriptions status messages also include the first (6) non-remote outputs.
```

What does it mean for an output the be "remote" anyway? And what's the opposite of that? "local"?

bosch_alarm_mode2/panel.py
sanjay900: It's actually to fix a limitation with the panel, if you toggle multiple outputs at once it will just not send subscription events without the semaphore

bosch_alarm_mode2/panel.py
sanjay900: You can set an outputs type, for example you may have a siren or something. Remote is just another one of these types but bizarrely only the remote ones are controllable via mode 2.

bosch_alarm_mode2/panel.py
sanjay900: I have asked bosch about this, but for now its best to leave it there, otherwise we end up in a state when home assistant thinks an output is in on state when its in a different state on the panel itself.

bosch_alarm_mode2/panel.py
```diff
@@ -382,43 +420,48 @@ async def _load_names_cf03(self, name_cmd, names) -> dict[int, str]:
...
         if self._supports_command_request_area_text_cf03:
-            return await self._load_names_cf03(name_cmd, names)
+            return await self._load_names_cf03(name_cmd, enabled_ids)
 
         if self._supports_command_request_area_text_cf01:
-            return await self._load_names_cf01(name_cmd, names)
+            return await self._load_names_cf01(name_cmd, enabled_ids, id_size)
+        
+        # And then if CF01 isn't available, we can just generate a list of names and return that
+        names = {}
```
mag1024, line 461: return { id:f"..." for id in enabled_ids }

bosch_alarm_mode2/panel.py
mag1024: Ok, let's add a comment documenting the reason this is needed, please

It is a bit suspect though, because even without a semaphore, we're not actually triggering the outputs "at the same time", but sending two separate commands. This makes me think that what is actually needed here is some kind of a delay/rate limit, not mutual exclusion, and a semaphore just happens to provide enough of a delay that it works.

bosch_alarm_mode2/panel.py
```diff
@@ -334,6 +363,10 @@ async def _basicinfo(self):
         if data[0] <= 0x24:
             self._partial_arming_id = AREA_ARMING_STAY1
             self._all_arming_id = AREA_ARMING_AWAY
+            # The solution panels only offer control over outputs with the "remote output" type.
+            # For most commands, output 0 is the first remote output.
+            # However, subscriptions status messages include information about all outputs, and remote outputs start at index 6.
```
mag1024, line 368: nit: wrap this line please

mag1024 (approved):

bosch_alarm_mode2/panel.py
sanjay900: You are probably correct about that, but this seems to work well enough and i don't suspect people will need to turn multiple outputs on simultaneously often, so this is probably good enough for a solution for now.

mag1024:
Thanks!

mag1024:
Just ran this on my panel, and I get this:
{1: : Inactive, 2: : Inactive, 3: : Inactive, 4: : Inactive, 5: : Inactive, 6: : Inactive, 7: : Inactive, 8: : Inactive, 9: : Inactive, 10: : Inactive, 11: : Inactive, 12: : Inactive, 13: : Inactive, 14: : Inactive, 15: : Inactive, 16: : Inactive, 17: : Inactive, 18: : Inactive, 19: : Inactive, 20: : Inactive, 21: : Inactive, 22: : Inactive, 23: : Inactive, 24: : Inactive, 25: : Inactive, 26: : Inactive, 27: : Inactive, 28: : Inactive, 29: : Inactive, 30: : Inactive, 31: : Inactive, 32: : Inactive, 33: : Inactive, 34: : Inactive, 35: : Inactive, 36: : Inactive, 37: : Inactive, 38: : Inactive, 39: : Inactive, 40: : Inactive, 41: : Inactive, 42: : Inactive, 43: : Inactive, 44: : Inactive, 45: : Inactive, 46: : Inactive, 47: : Inactive, 48: : Inactive, 49: : Inactive, 50: : Inactive, 51: : Inactive, 52: : Inactive, 53: : Inactive, 54: : Inactive, 55: : Inactive, 56: : Inactive, 57: : Inactive, 58: : Inactive, 59: : Inactive, 60: : Inactive, 61: : Inactive, 62: : Inactive, 63: : Inactive, 64: : Inactive, 253: RELAY A: Inactive, 254: RELAY B: Inactive, 255: RELAY C: Inactive}

I wonder if the unnamed outputs are ever useful...

mag1024:
BTW, I installed RPS again to try to see what outputs the panel thinks it has configured, but I get the "Invalid or missing hardware key. Communication unavailable".

sanjay900:
Thats interesting, even my B series panel only shows me outputs i have enabled

sanjay900:
One slightly unfortunate thing is i started playing with outputs more, and worked out that that offset for solution series panels isn't actually totally correct, but i'm not actually sure how to make it work correctly so i have reached out to bosch to see if i can find some command to get the correct offsets.

mag1024:
It could be that those first 64 "unnamed" outputs are in the same category of "non-remote" as the first 6 outputs on the Solutions panels. It's hard for me to confirm what's going on without programming access to my panel -- is that something your Bosch contacts might be able to help with? It still seems like it may require some dongle, but maybe I'm just doing something wrong, and they can give me better instructions.
 
In the meantime, should we disable creation of output entities without a name? (This could be done either in the library, or the integration) Otherwise I've been holding off on updating my main instance because I don't want to manually go through and disable so many spurious entities.

sanjay900:
I think thats totally reasonable. Which panel have you got again?

sanjay900:
@mag1024 would you be able to give me some contact details? I can forward them on to my bosch contacts and they can put you in contact with someone who can walk you through configuring your panel.

mag1024:
B8512G.
Whenever I've tried to use RPS to connect to the panel, I get the "Invalid or missing hardware key. Communication unavailable" -- regardless of whether I use the local or the external IP.

sanjay900:
I have flicked your details across.

sanjay900:
https://github.com/mag1024/bosch-alarm-mode2/pull/21
and here is the PR to skip the outputs with empty names.

mag1024:
> @mag1024 Oh yeah FYI i did learn one thing the other week you will probably find useful when testing mode2 stuff, you can just clone it to your home assistant config directory, and it will use that version of the python module.

Got this to work! One note for future generations is that it is the library directory, not the root of the repo that has to be in /config. Easily solved with `ln -s bosch-alarm-mode2/bosch_alarm_mode2/ .`

mag1024:
> I have flicked your details across.

Thank you so much! They got in touch, and indeed the "large" panels require some special USB dongle to use RPS. They may be able to hook me up.

sanjay900:
> > I have flicked your details across.
> 
> Thank you so much! They got in touch, and indeed the "large" panels require some special USB dongle to use RPS. They may be able to hook me up.

Nice! Yeah they are pretty good with providing hardware, I should be receiving an AMAX panel and some output expansion modules soon.

PR #19, opened 2023-10-12
Use UTF-8 encoding for area text names

ocitrev:
Add a configurable encoding for names.  My B4512 panel has names encoded with utf8 and decoding ascii fails with utf8 bytes


ocitrev:
I was getting error
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 3: ordinal not in range(128)
```

sanjay900:
I think it would just make more sense to always use utf-8. No real gain to making it configurable.

ocitrev:
Allright, I'll make it utf8 then.

PR #20, opened 2023-10-12
Fallback to CF01 when CF03 fails

ocitrev:
Not sure about this patch ...

I don't know why; I can't load names from my B4512 panel with CF03 even when support should be enabled.
`_load_names_cf03` receives a truncated response.  Reverting to CF01 did the trick for me.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -453,7 +453,11 @@ class Panel:
         enabled_ids = await self._load_entity_set(config_cmd)
         
         if self._supports_command_request_area_text_cf03:
-            return await self._load_names_cf03(name_cmd, enabled_ids)
+            try:
+                return await self._load_names_cf03(name_cmd, enabled_ids)
+            except:
+                # If the panel doesn't really support CF03, then use CF01
+                self._supports_command_request_area_text_cf03 = False
 
         if self._supports_command_request_area_text_cf01:
             return await self._load_names_cf01(name_cmd, enabled_ids, id_size)
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -456,8 +456,11 @@ class Panel:
             try:
                 return await self._load_names_cf03(name_cmd, enabled_ids)
             except:
-                # If the panel doesn't really support CF03, then use CF01
-                self._supports_command_request_area_text_cf03 = False
+                if self._connection.protocol == PROTOCOL.BASIC:
+                    raise
+                # downgrade protocol to basic and retry
+                self._connection.protocol = PROTOCOL.BASIC
+                return await self._load_names_cf03(name_cmd, enabled_ids)
 
         if self._supports_command_request_area_text_cf01:
             return await self._load_names_cf01(name_cmd, enabled_ids, id_size)


EHylands:
> Not sure about this patch ...
> 
> 
> 
> I don't know why; I can't load names from my B4512 panel with CF03 even when support should be enabled.
> 
> `_load_names_cf03` receives a truncated response.  Reverting to CF01 did the trick for me.
> 
> 

Not directly related with this code.
Also tried using CF03 commands on my B6512 panel with extended 0x04 protocol and also got truncated responses ...

ocitrev:
I got error
```
Traceback (most recent call last):
  File "bosch-alarm-mode2\bin\alarm.py", line 29, in <module>
    loop.run_until_complete(panel.connect())
  File "Python\Python311\Lib\asyncio\base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "bosch-alarm-mode2\bin\..\bosch_alarm_mode2\panel.py", line 161, in connect
    await self._connect(load_selector)
  File "bosch-alarm-mode2\bin\..\bosch_alarm_mode2\panel.py", line 248, in _connect
  File "bosch-alarm-mode2\bin\..\bosch_alarm_mode2\panel.py", line 168, in load
    await self._load_points()
  File "bosch-alarm-mode2\bin\..\bosch_alarm_mode2\panel.py", line 410, in _load_points
    names = await self._load_names(CMD.POINT_TEXT, CMD.REQUEST_CONFIGURED_POINTS, "POINT")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "bosch-alarm-mode2\bin\..\bosch_alarm_mode2\panel.py", line 457, in _load_names
    return await self._load_names_cf03(name_cmd, enabled_ids)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "bosch-alarm-mode2\bin\..\bosch_alarm_mode2\panel.py", line 424, in _load_names_cf03
    name, data = data[2:].split(b'\x00', 1)
    ^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)
```

ocitrev:
> Not directly related with this code. Also tried using CF03 commands on my B6512 panel with extended 0x04 protocol and also got truncated responses ...

The response I got is only 96 bytes long.  Is there a limit on the length of the data returned ?

EHylands:
> > Not directly related with this code. Also tried using CF03 commands on my B6512 panel with extended 0x04 protocol and also got truncated responses ...
> 
> 
> 
> The truncated response I got is only 96 bytes long.  Is there a limit on the length of the data returned?

Extended protocol 0x04 allows for 1460 bytes commands and responses.

For testing purpose, have you tried reverting back to Basic protocol 0x01 ?

 if bitmask[0] & 0x10:
            self._connection.protocol = PROTOCOL.EXTENDED

sanjay900:
I have a B5512 myself, so ill configure a bunch of points and replicate this one, and see if i can find a nicer fix than this.

EHylands:
> I have a B5512 myself, so ill configure a bunch of points and replicate this one, and see if i can find a nicer fix than this, or at least work out why its truncating.
> 
> How many points have you got configured?

I'm running  27 Points and 2 Areas

Did further testing:
Mode2ReqAreaText CF03:  works well with protocol 0x04.

Mode2ReqPointText CF03: I get truncated and incomplete responses from panel
Sending : [ 04, 00, 05, 3C, 00, 00, 00,  01]
Receiving: [04 00 08 fe 00 01 50 6f 72 74 65] (packet well formed, but incomplete data ... )

sanjay900:
Ive asked my bosch contact if they have any idea why this happens, the question has been forwarded to the engineers.

ocitrev:
Would it make more sense to revert to basic protocol then?
I tried this code below and it works fine on my B4512 panel:

```python
if self._supports_command_request_area_text_cf03:
    try:
        return await self._load_names_cf03(name_cmd, enabled_ids)
    except:
        self._connection.protocol = PROTOCOL.BASIC
        return await self._load_names_cf03(name_cmd, enabled_ids)
```

EHylands:
> Ive asked my bosch contact if they have any idea why this happens, the question has been forwarded to the engineers.

Did you report to Bosch AU ou US ?

ocitrev:
> How many points have you got configured?

I have 18 points and 1 area

justin-boschsecurity:
> > I have a B5512 myself, so ill configure a bunch of points and replicate this one, and see if i can find a nicer fix than this, or at least work out why its truncating.
> > How many points have you got configured?
> 
> I'm running 27 Points and 2 Areas
> 
> Did further testing: Mode2ReqAreaText CF03: works well with protocol 0x04.
> 
> Mode2ReqPointText CF03: I get truncated and incomplete responses from panel Sending : [ 04, 00, 05, 3C, 00, 00, 00, 01] Receiving: [04 00 08 fe 00 01 50 6f 72 74 65] (packet well formed, but incomplete data ... )

What firmware version is your panel currently running?
Are you using the on-board Ethernet adapter or are you using an add-on B426 Ethernet adapter?
Can you take a screenshot of the Point Assignments section of RPS at least showing the points' text?
You can also reach out to integrated.solutions@us.bosch.com for more help with the protocol itself.

ocitrev:
> What firmware version is your panel currently running?

I updated to version 3.14.0.12 when it was not working with the original firmware version and got the same issue.

> Are you using the on-board Ethernet adapter or are you using an add-on B426 Ethernet adapter?

On-board Ethernet adapter

> Can you take a screenshot of the Point Assignments section of RPS at least showing the points' text?

image

Better yet, here is the truncated data with protocol 0x04 then the retry with protocol 0x01

```
DEBUG: >> b'\x04\x00\x05<\x00\x00\x00\x01'
DEBUG: << b'\x04\x00]\xfe\x00\x01Porte Avant\x00\x00\x02Mouvement Cuisine\x00\x00\x03Mouvement Salon\x00\x00\x04Porte Patio\x00\x00\x05Porte Arriere Garage\x00\x00\x06P'
DEBUG: >> b'\x01\x05<\x00\x00\x00\x01'
DEBUG: << b'\x01\xc2\xfe\x00\x01Porte Avant\x00\x00\x02Mouvement Cuisine\x00\x00\x03Mouvement Salon\x00\x00\x04Porte Patio\x00\x00\x05Porte Arriere Garage\x00\x00\x06Porte Maison/Garage\x00\x00\x07Mouvement Sous-Sol\x00\x00\x08Fenetre Salon Sous-Sol\x00\x00\tMouvement Garage\x00\x00\nFenetre Garage\x00'
```

EHylands:
> > > I have a B5512 myself, so ill configure a bunch of points and replicate this one, and see if i can find a nicer fix than this, or at least work out why its truncating.
> > > How many points have you got configured?
> > 
> > 
> > I'm running 27 Points and 2 Areas
> > Did further testing: Mode2ReqAreaText CF03: works well with protocol 0x04.
> > Mode2ReqPointText CF03: I get truncated and incomplete responses from panel Sending : [ 04, 00, 05, 3C, 00, 00, 00, 01] Receiving: [04 00 08 fe 00 01 50 6f 72 74 65] (packet well formed, but incomplete data ... )
> 
> What firmware version is your panel currently running? Are you using the on-board Ethernet adapter or are you using an add-on B426 Ethernet adapter? Can you take a screenshot of the Point Assignments section of RPS at least showing the points' text? You can also reach out to [integrated.solutions@us.bosch.com](mailto:integrated.solutions@us.bosch.com) for more help with the protocol itself.

Same thing on my side
B6512
On board adapter 
3.12 firmware

@ocitrev I think Mode2ReqPointText CF03 doesn't work well with french PointText :P

Screenshot 2023-10-12 at 4 18 43 PM

mag1024:
Random observation: the packet seems to get cutoff on a point name that happens to be the only point that contains a special character ('/'). Might very well be a coincidence, but could you try renaming the point to remove it?

Regarding the PR: I actually liked the previous version -- the one that was falling back on cf01 -- better. Changing a global protocol version inside the handling of a particular command seems rather indirect.

EHylands:
> Random observation: the packet seems to get cutoff on a point name that happens to be the only point that contains a special character ('/'). Might very well be a coincidence, but could you try renaming the point to remove it?
> 
> Regarding the PR: I actually liked the previous version -- the one that was falling back on cf01 -- better. Changing a global protocol version inside the handling of a particular command seems rather indirect.

I tried removing any non alpha numeric character in my point text with same result

justin-boschsecurity:
> @ocitrev I think Mode2ReqPointText CF03 doesn't work well with french PointText :P

I was also concerned that it might be something about French text (was really thinking special characters) but I have copied your text into my B5512 and so far have not been able to reproduce the problem.

>I tried removing any non alpha numeric character in my point text with same result

I tried _adding in_ some non-alphanumeric characters in my own point text and still couldn't reproduce the problem.

I have a B6512 in my other office and can try that too, but they should be the same.

ocitrev:
> Regarding the PR: I actually liked the previous version -- the one that was falling back on cf01 -- better. Changing a global protocol version inside the handling of a particular command seems rather indirect.

I opted to downgrade the protocol from extended to basic, since CF03 works fine with protocol 0x01 and CF03 is much faster.

My original PR did set `_supports_command_request_area_text_cf03 = False` which is also an indirect change of a global variable.

In any case I don't really care how it is handled, I just want it to work ;)

sanjay900:
Do any of the other command formats work any better?

sanjay900:
Im curious what panels the extended protocol actually work correctly with at the moment. It might just end up making more sense to disable using the extended protocol at all if none of the panels fully support it.

sanjay900:
> Random observation: the packet seems to get cutoff on a point name that happens to be the only point that contains a special character ('/'). Might very well be a coincidence, but could you try renaming the point to remove it?
> 
> Regarding the PR: I actually liked the previous version -- the one that was falling back on cf01 -- better. Changing a global protocol version inside the handling of a particular command seems rather indirect.

Changing the protocol version here for sure is rather indirect - if we were going to go that route, we would have to work out what situations we need to downgrade the protocol and just never upgrade to extended for those panels.

mag1024:
> Im curious what panels the extended protocol actually work correctly with at the moment. It might just end up making more sense to disable using the extended protocol at all if none of the panels fully support it.

AFAICT, B8512G supports it without issues.
If we wanted to have an explicit whitelist/blacklist for enabling it on top of what the panel reports, that seems fine.

sanjay900:
> > Im curious what panels the extended protocol actually work correctly with at the moment. It might just end up making more sense to disable using the extended protocol at all if none of the panels fully support it.
> 
> AFAICT, B8512G supports it without issues. If we wanted to have an explicit whitelist/blacklist for enabling it on top of what the panel reports, that seems fine -- but we should do it up-front at initialization time.

For sure, ill check with my B5512 tonight and see what it does.

justin-boschsecurity:
>Im curious what panels the extended protocol actually work correctly with at the moment. It might just end up making more sense to disable using the extended protocol at all if none of the panels fully support it.
The extended protocol was created for the US based B/G Series panels so it _should_ work in all of them (that does not include the GV4 panels). I'm pretty sure the Bosch Security Manager app uses the extended protocol with them all as it is so this _shouldn't_ be an issue. If I can find a way to duplicate the problem, I will raise the issue with engineering to find the root cause and get it fixed.

justin-boschsecurity:
I was just able to duplicate the behavior after adding a few more configured points to my panel.  FWIW, it cut off in the middle of point 7's text for me. 
I'll try some debugging in the panel and see what I can come up with.

justin-boschsecurity:
Okay, I see what the problem is. Everything works fine as long as the length of the returned text is 256 bytes or lower. As soon as it crosses over to 257 the buffer starts over at just one byte. 
Note that a big difference between the Extended Protocol 0x04 and the standard Command and Response protocol 0x01 is that the former has a two-byte (16 bit) length parameter while the latter has a one-byte (8-bit) length parameter.
Most likely this is a bug in the firmware that improperly calculates that length value and ends up truncating the returned value.

I will report this to the panel engineers but the best we can hope for is a fix in a future release. The current version is 3.14 and it's been out for a while so there's a good chance we won't get the fix in before 3.15 is released but I think it's reasonable to expect it to be fixed by 3.16.

Of course, that doesn't help this project much aside from knowing it is a panel firmware bug and you should treat any US panels as not supporting the extended version of the point text command.

Here's a full list of the models that would be affected by this bug; B3512, B4512, B5512, B6512, B8512G, B9512G

sanjay900:
In that case, lets maybe just disable the extended protocol for the moment, and we can re-evaluate once a new firmware is out with a fix.

ocitrev:
I agree. I'll close this PR since not needed anymore.
Thanks !

mag1024:
> In that case, lets maybe just disable the extended protocol for the moment, and we can re-evaluate once a new firmware is out with a fix.

Yep, sounds good.

PR #21, opened 2023-10-12
Ignore outputs with an empty name

sanjay900:
Hopefully we will find a better solution than this, but I can confirm that if i leave the name empty on my outputs, they just use the default OUTPUTx name.


mag1024:
Thank you!

sanjay900:
@mag1024 Sadly, i just tested this one again and found that this fix actually doesn't work on my solution 3000, so we may need to revert it, and then figure out why it doesn't work on your panel.

PR #22, opened 2023-10-12
Disable the extended protocol for now

sanjay900:
As noted in https://github.com/mag1024/bosch-alarm-mode2/pull/20, there is currently a bug in the firmwares that support this protocol, and I don't think it provides all that much value currently as we don't pull back lots of history anymore. In the future we may be able to enable this again once a fixed firmware is released.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -374,8 +374,8 @@ class Panel:
             self._all_arming_id = AREA_ARMING_MASTER_DELAY
         # Section 13.2 of the protocol spec.
         bitmask = data[23:].ljust(33, b'\0')
-        if bitmask[0] & 0x10:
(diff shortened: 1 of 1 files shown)


bosch_alarm_mode2/panel.py
```diff
@@ -374,8 +374,8 @@ async def _basicinfo(self):
             self._all_arming_id = AREA_ARMING_MASTER_DELAY
         # Section 13.2 of the protocol spec.
         bitmask = data[23:].ljust(33, b'\0')
-        if bitmask[0] & 0x10:
-            self._connection.protocol = PROTOCOL.EXTENDED
+        # if bitmask[0] & 0x10:
```
mag1024, line 377: Let's leave a comment describing the bug, or linking to the issue where it is discussed -- particularly let's note when we should revisit this.

PR #23, opened 2023-10-19
AMAX Support

sanjay900:
Now that i have my hands on a AMAX 2100 panel, I can begin getting this working

One super annoying thing i have found is that the auth process for these panels is doing both automation and remote user, which means we need to change the integration to allow specifying a code, a passcode or both.


sanjay900:
I am also having some issues with the history ID which i will have to try to figure out a way around, and the `request alarm memory summary` command isn't supported here, so we will have to find another way to work out what alarms are triggered.

sanjay900:
Seems like my AMAX panel just handles its history event IDs totally differently. With this panel, a request for event 0 returns the latest events, and trying to retrieve more events than the panel has just wraps around to the beginning instead of returning no events. This would mean we would need to change the logic a touch, to instead always fetch from event 0 on these panels, and then only append new events. This is a bit messy though!

mag1024:
Ooh, yeah, that's kinda messy... Seems like this implementation is quite different. Is Amax Bosch's oldest panel?
If possible, let's please keep the auth changes, the command formats, and the history changes all separate from each other for PR purposes.

sanjay900:
I think I'm just trying to get things working first and then I can split the changes.

I think I understand what's happening now: the AMAX panels only store a max of 255 events and then it rolls over, and whatever logic they use for that seems to not handle things quite right. I do wonder if there's a better solution to fix this stuff.

sanjay900:
https://github.com/mag1024/bosch-alarm-mode2/pull/25
https://github.com/mag1024/bosch-alarm-mode2/pull/24
Both of these will fix everything besides history, i will figure out how the hell to resolve the history problem in another branch once i have that one worked out.

PR #24, opened 2023-10-20
Check commands for compatibility before using them

sanjay900:
AMAX panels don't support `Request output text` and more strangely, they don't support alarm memory summary CF01 but do support CF02.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -15,6 +15,12 @@ ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
 ssl_context.check_hostname = False
 ssl_context.verify_mode = ssl.CERT_NONE
 ssl_context.set_ciphers('DEFAULT')
+
+def _supported_format(value, masks):
+    for mask, format in masks:
+        if value & mask:
+            return format
+    return 0
 class PanelEntity:
     def __init__(self, name, status):
         self.name = name
@@ -350,12 +356,6 @@ class Panel:
             await self._authenticate_automation_user()
         else:
             await self._authenticate_remote_user()
-    
-    def _supported_format(self, value, masks):
-        for mask, format in masks:
-            if value & mask:
-                return format
-        return 0
 
     async def _basicinfo(self):
         try:
@@ -392,10 +392,10 @@ class Panel:
         self._supports_serial = bitmask[13] & 0x04
         self._supports_status = bitmask[5] & 0x08
         self._supports_subscriptions = bitmask[0] & 0x40
-        self._area_text_supported_format = self._supported_format(bitmask[7], ((0x08, 3), (0x20, 1)))
-        self._output_text_supported_format = self._supported_format(bitmask[9], ((0x10, 3), (0x40, 1)))
-        self._point_text_supported_format = self._supported_format(bitmask[11], ((0x20, 3), (0x80, 1)))
-        self._alarm_summary_supported_format = self._supported_format(bitmask[2], ((0x10, 2), (0x20, 1)))
+        self._area_text_supported_format = _supported_format(bitmask[7], [(0x08, 3), (0x20, 1)])
+        self._output_text_supported_format = _supported_format(bitmask[9], [(0x10, 3), (0x40, 1)])
+        self._point_text_supported_format = _supported_format(bitmask[11], [(0x20, 3), (0x80, 1)])
+        self._alarm_summary_supported_format = _supported_format(bitmask[2], [(0x10, 2), (0x20, 1)])
 
         self._history.init_for_panel(data[0])
         self._history_cmd = (
@@ -507,7 +507,7 @@ class Panel:
                 LOG.warning(
                     f"Found unknown area {area}, supported areas: [{self.areas.keys()}]")
             response_detail = response_detail[5:]
-        
+
     async def _get_alarm_status(self):
         format = None
         if not self._alarm_summary_supported_format:
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -143,14 +143,10 @@ class Panel:
         self._all_arming_id = AREA_ARMING_MASTER_DELAY
         self._supports_serial = False
         self._supports_subscriptions = False
-        self._supports_command_request_area_text_cf01 = False
-        self._supports_command_request_area_text_cf03 = False
-        self._supports_command_request_output_text_cf01 = False
-        self._supports_command_request_output_text_cf03 = False
-        self._supports_command_request_point_text_cf01 = False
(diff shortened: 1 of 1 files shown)


bosch_alarm_mode2/panel.py
```diff
@@ -403,15 +415,30 @@ async def _extended_info(self):
             self.firmware_version = 'v%d.%d' % (version, revision)
 
     async def _load_outputs(self):
-        names = await self._load_names(CMD.OUTPUT_TEXT, CMD.REQUEST_CONFIGURED_OUTPUTS, "OUTPUT", 1)
+        names = await self._load_names(
+            CMD.OUTPUT_TEXT, CMD.REQUEST_CONFIGURED_OUTPUTS, 
+            self._supports_command_request_output_text_cf01,
```
mag1024, line 420: The number of bools at the top level is getting a bit out of control :)

How about we turn all the parameters to load_names into a tuple, store them as member variables, and update the "supports" directly in them?

So instead of `self._supports_command_request_point_text_cf01 = bitmask[11] & 0x80` it'll become `self._load_points_spec.supports_cf01 = bitmask[11] & 0x80`

And _load_points() will become just `await self._load_names(self._load_points_spec)`

Or, alternatively/additionally, we can have turn the supported formats into a set, so one can write `1 in point_text_supported_formats`

mag1024 (changes requested):

bosch_alarm_mode2/panel.py
sanjay900: thinking about it now, we don't even need to go as extreme as a tuple. We can just store the highest supported command, and then compare with that.

sanjay900:
Is there a more pythonic way to write these sections?
```python
 if bitmask[7] & 0x08:
      self._area_text_supported_format = 3
 elif bitmask[7] & 0x20:
      self._area_text_supported_format = 1
  ```

File: bosch_alarm_mode2/panel.py [hunk 2]
```diff
@@ -384,8 +386,27 @@ async def _basicinfo(self):
         self._supports_serial = bitmask[13] & 0x04
         self._supports_status = bitmask[5] & 0x08
         self._supports_subscriptions = bitmask[0] & 0x40
-        self._supports_command_request_area_text_cf01 = bitmask[7] & 0x20
-        self._supports_command_request_area_text_cf03 = bitmask[7] & 0x08
+
+        if bitmask[7] & 0x08:
+            self._area_text_supported_format = 3
```
mag1024, line 391: You could do a nested ternary (x if c else y), but for the sake of readability perhaps a helper function is the way to go

bosch_alarm_mode2/panel.py
```diff
@@ -454,13 +487,13 @@ async def _load_entity_set(self, cmd) -> [int]:
             index += 8
         return ids
 
-    async def _load_names(self, name_cmd, config_cmd, type, id_size=2) -> dict[int, str]:
+    async def _load_names(self, name_cmd, config_cmd, supported_format, type, id_size=2) -> dict[int, str]:
         enabled_ids = await self._load_entity_set(config_cmd)
         
-        if self._supports_command_request_area_text_cf03:
+        if supported_format==3:
```
mag1024, line 493: here and below: spaces around ==, please

bosch_alarm_mode2/panel.py
```diff
@@ -484,9 +517,16 @@ async def _get_alarms_for_priority(self, priority, last_area=None, last_point=No
                 LOG.warning(
                     f"Found unknown area {area}, supported areas: [{self.areas.keys()}]")
             response_detail = response_detail[5:]
-
+        
     async def _get_alarm_status(self):
-        data = await self._connection.send_command(CMD.ALARM_MEMORY_SUMMARY)
+        format = None
```
mag1024, line 522: Consider:

```
if not self._alarm_summary_supported_format:
  return
format = bytearray([0x02] if self._alarm_summary_supported_format == 2 else [])
```

bosch_alarm_mode2/panel.py
```diff
@@ -143,8 +143,10 @@ def __init__(self, host, port, passcode):
         self._all_arming_id = AREA_ARMING_MASTER_DELAY
         self._supports_serial = False
         self._supports_subscriptions = False
-        self._supports_command_request_area_text_cf01 = False
-        self._supports_command_request_area_text_cf03 = False
+        self._area_text_supported_format = 0
```
mag1024, line 146: Nice, I like this approach a lot!

In the future maybe we can still add some better organization by introducing a helper class named something like PanelFeatures, and moving all these members, as well as the logic for populating them there.

mag1024 (changes requested):

bosch_alarm_mode2/panel.py
sanjay900: Yeah for sure that would make sense if we had more of these

bosch_alarm_mode2/panel.py
mag1024: The function I was thinking of would be something like
`self._area_text_supported_format = _supported_format(bitmask[7], [(0x08, 3), (0x20, 1)])

bosch_alarm_mode2/panel.py
```diff
@@ -348,6 +350,12 @@ async def _authenticate(self):
             await self._authenticate_automation_user()
         else:
             await self._authenticate_remote_user()
+    
+    def _supported_format(self, value, masks):
```
mag1024, line 354: Doesn't have to be be a member of Panel -- let's move it out

bosch_alarm_mode2/panel.py
```diff
@@ -384,8 +392,11 @@ async def _basicinfo(self):
         self._supports_serial = bitmask[13] & 0x04
         self._supports_status = bitmask[5] & 0x08
         self._supports_subscriptions = bitmask[0] & 0x40
-        self._supports_command_request_area_text_cf01 = bitmask[7] & 0x20
-        self._supports_command_request_area_text_cf03 = bitmask[7] & 0x08
+        self._area_text_supported_format = self._supported_format(bitmask[7], ((0x08, 3), (0x20, 1)))
```
mag1024, line 395: not sure if it makes a practical difference, but I think a list would make more sense than a tuple for the outer container

bosch_alarm_mode2/panel.py
```diff
@@ -484,9 +507,13 @@ async def _get_alarms_for_priority(self, priority, last_area=None, last_point=No
                 LOG.warning(
                     f"Found unknown area {area}, supported areas: [{self.areas.keys()}]")
             response_detail = response_detail[5:]
-
+
```
mag1024, line 510: nit: extraneous spaces

mag1024:
Thank you!

PR #25, opened 2023-10-20
Implement auth for AMAX panels

sanjay900:
The AMAX panels require both an automation code based auth and a remote user based auth. This means we need to support passing in data for both types of auth.


sanjay900:
From a home assistant end we will need to change the config format, and it would be best to migrate any currently stored codes to both variables where it makes sense, so that solution users don't need to reconfigure their integrations.

sanjay900:
Looking into some solution stuff, it may pay to require both for solution panels as well, as it appears the only way to sort the output problems for the solution panels will be to read and parse the configuration from the panel, which requires being logged in with both codes.

bosch_alarm_mode2/panel.py
```diff
@@ -359,6 +360,12 @@ async def _basicinfo(self):
         self.protocol_version = 'v%d.%d' % (data[5], data[6])
         if data[13]:
             LOG.warning('busy flag: %d', data[13])
+        # Solution series uses only remote user 
+        # Amax series uses remote user + automation code
+        # Everything else uses just automation code
+        self._supports_remote_user = data[0] <= 0x24
```
mag1024, line 366: If we're going to have bools, they should probably be intialized in __init__ -- but maybe we can null out the passcode/automation_code as appropriate instead, and avoid the bools altogether?

mag1024 (changes requested):

bosch_alarm_mode2/panel.py
sanjay900: Seems reasonable, i will just require both codes for the solution panels as well, so i will just null out the user code on b/g series panels.

sanjay900:
One thing to note is that this will change authentication for solution panels as well, but that just means we need to make sure we warn users about that when they get the update.

README.md
```diff
@@ -10,8 +10,9 @@ In practice, this library has only been tested with B8512G and the B5512 and the
 - Push based updates (for panels that support it)
 
 #### Authentication
-- For B and G series panels, in the "Automation / Remote App" section of panel settings set the "Automation Device" to "Mode 2", and set the "Automation Passcode" to at least 10 characters.
-- For the Solution panels, use your user code
+- For all panels, make sure that your Automation Passcode is set to a passcode that is at least 10 characters long.
+- For B and G series panels, set the "Automation Device" to "Mode 2", and use just your automation code for authentication.
+- For Solution and AMAX panels, use both your automation code and your installer code for authentication.
```
mag1024, line 15: So "installer" is the like root user of the panel?
Would another user's code theoretically work here -- possibly only showing the entities the user has permission for?
I'm trying to confirm the terminology, because if it's actually the case that the _installer_ code is required, we should probably rename the `passcode` member/arugment in the Panel to `installer_code`

bin/alarm.py
```diff
@@ -10,6 +10,7 @@
 cli_parser = argparse.ArgumentParser()
 cli_parser.add_argument('--host', help="panel hostname")
 cli_parser.add_argument('--port', type=int, help="panel port")
+cli_parser.add_argument('-U', '--installercode', help="Installer code")
```
mag1024, line 13: posix flag naming convention would suggest --installer-code; but this is conditional on the answer to the question above

README.md
sanjay900: Sadly for the solution panels and amax panels it does need to be an installer code

sanjay900: Non installer codes won't give access to the commands we need for the solution panel, and non installer codes don't seem to work on the amax panels

bin/alarm.py
sanjay900: Makes sense, I'll change that

bosch_alarm_mode2/panel.py
```diff
@@ -117,11 +117,12 @@ def __repr__(self):
 class Panel:
     """ Connection to a Bosch Alarm Panel using the "Mode 2" API. """
 
-    def __init__(self, host, port, passcode):
+    def __init__(self, host, port, installer_code, automation_code):
```
mag1024, line 120: nit: let's put `automation_code`, as the non-optional parameter, before the optional `installer_code`

bin/alarm.py
```diff
@@ -10,6 +10,7 @@
 cli_parser = argparse.ArgumentParser()
 cli_parser.add_argument('--host', help="panel hostname")
 cli_parser.add_argument('--port', type=int, help="panel port")
+cli_parser.add_argument('-U', '--installer-code', help="Installer code")
 cli_parser.add_argument('-P', '--passcode', help="Automation passcode")
```
mag1024, line 14: For full consistency let's also rename this to --automation-code

bosch_alarm_mode2/panel.py
```diff
@@ -368,10 +371,11 @@ async def _basicinfo(self):
             # However, subscriptions status messages include information about all outputs. 
             # Outputs with the "remote output" type start at index 6.
             self._output_subscription_start_index = 6
```
mag1024, line 373: Should we be checking that the _installer_code is non-empty on this branch?

PR #26, opened 2023-10-20
Fix history for AMAX panels

sanjay900:
Turns out my panel was just in a busted state, after arming and disarming my panel it now returns  data in a way that actually makes sense, however AMAX panels do use the first byte to signal different event types, so we need to mask that away when we get an event ID  from the panel.

### DIFF
--- a/bosch_alarm_mode2/history.py
+++ b/bosch_alarm_mode2/history.py
@@ -25,7 +25,6 @@ class History:
         self._events = []
         self._parser = None
         self._max_count = 0
-        self._event_id_mask = 0xFFFFFFFF
 
     @property
     def events(self):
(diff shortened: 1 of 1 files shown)


bosch_alarm_mode2/history.py
```diff
@@ -42,6 +43,9 @@ def init_for_panel(self, panel_type):
             self._parser = SolutionHistoryParser()
         elif panel_type <= 0x24:
             self._parser = AmaxHistoryParser()
+            # AMAX panels use the first byte of the event id as a flag for the event type
+            # We need to mask that away to get a proper event id
+            self._event_id_mask = 0x00FFFFFF
```
mag1024, line 48: It's a bit hard to believe that any panel would have more than 6 bytes-worth of event ids -- can we just apply this mask unconditionally?

sanjay900: Seems reasonable

PR #27, opened 2023-10-23
Fix output mapping for solution panels

sanjay900:
Solution panels are very weird with how outputs work. 

There is a concept of "Remote outputs", which govern which outputs can be controlled by Mode2. 
This gives the installer control over which outputs map to a given remote output that is controlled by Mode2. 

The problem with this, is that subscription events give us information based on the raw output ID, not the remote one.

To solve this, I have parsed the raw configuration of the panel, and used that to build a mapping from raw output ID to remote output ID. I then use this mapping when we receive a subscription event to make sure the event IDs are aligned.

Do note that this requires we merge https://github.com/mag1024/bosch-alarm-mode2/pull/25 first, as these configuration commands are only available to installers, and https://github.com/mag1024/bosch-alarm-mode2/pull/25 changes the user scope from user to installer.

Also note that I have  undone https://github.com/mag1024/bosch-alarm-mode2/pull/21 here as well, as when testing this I found that outputs without a name are coming through as empty from my solution 3000, which means empty but configured outputs were being ignored. We will need to work out why `REQUEST_CONFIGURED_OUTPUTS` does not work on your panel? Were you able to get your hands on a configuration key so you can validate your configuration, and check if those empty output configurations are in fact valid?

### DIFF
--- a/bosch_alarm_mode2/const.py
+++ b/bosch_alarm_mode2/const.py
@@ -177,6 +177,7 @@ class CMD:
     SET_SUBSCRIPTION = 0x5F
     # Diagnostic group
     PRODUCT_SERIAL = 0x4A
+    READ_CONFIG_IMAGE = 0xA0
 
 class PROTOCOL:
     BASIC = 0x01
--- a/bosch_alarm_mode2/const.py
+++ b/bosch_alarm_mode2/const.py
@@ -177,7 +177,6 @@ class CMD:
     SET_SUBSCRIPTION = 0x5F
     # Diagnostic group
     PRODUCT_SERIAL = 0x4A
-    READ_CONFIG_IMAGE = 0xA0
 
 class PROTOCOL:
     BASIC = 0x01
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -445,6 +445,7 @@ class Panel:
         return data_out
 
     async def _load_output_subscription_config(self):
+        # Construct a mapping from remote outputs to actual outputs on solution panels
         # Locations for various outputs in memory
         # https://csproducts.co.nz/download/20003000_manuals/2000-3000-Quick-Installer-Manual.pdf
         output_1_location = 436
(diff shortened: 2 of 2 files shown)


sanjay900:
I will admit I have just guessed all of this, as the configuration logic is not really documented anywhere, but this all does seem to line up and the logic is working on my panel.

sanjay900:
It has just occurred to me, that if we are parsing the config, we would be able to do more with that such as actually reading a points type directly from the config instead of guessing it by the name.

mag1024:
Oof, this is pretty gnarly!
 
The fact that subscriptions use some internal indexes that are different from how the outputs are addressed within the rest of the mode 2 protocol straight up sounds like a bug -- maybe @justin-boschsecurity can get the eng team to check?

How about we ignore the content of output subscription event entirely, and just use it to trigger a _load_output_status?

Haven't gotten the dongle yet - last I heard they were checking if there were any in the New York state office, or if one had to be sent from Australia. Let me ping them...

sanjay900:
Yeah, it's not a bug, they actually recommended me doing things like this.

I am happy to just trigger a load output status instead that probably would be a touch neater.

justin-boschsecurity:
Sorry, I can't help with the Solution Series. I only know B/G Series side.

bosch_alarm_mode2/const.py
```diff
@@ -177,6 +177,7 @@ class CMD:
     SET_SUBSCRIPTION = 0x5F
     # Diagnostic group
     PRODUCT_SERIAL = 0x4A
+    READ_CONFIG_IMAGE = 0xA0
```
mag1024, line 180: No longer used

bosch_alarm_mode2/panel.py
```diff
@@ -598,10 +594,10 @@ def _area_ready_consumer(self, data) -> int:
         return 5
 
     def _output_status_consumer(self, data) -> int:
-        output_id = BE_INT.int16(data) - self._output_subscription_start_index
-        if output_id in self.outputs:
-            output_status = self.outputs[output_id].status = int(data[2] != 0)
-            LOG.debug("Output updated %d: %s" % (output_id, OUTPUT_STATUS.TEXT[output_status]))
+        # Solution panels send us events with different output IDs.
```
mag1024, line 597: `# Solution panels send events with output ids that don't match those used by the rest of the commands.`

mag1024 (approved):
Cool, this is much nicer!

PR #28, opened 2023-10-31
Revert solution auth changes

sanjay900:
With recent changes to https://github.com/mag1024/bosch-alarm-mode2/pull/27, we no longer need installer code permissions on the solution panels. This reverts that change, so that solution panels instead just use the standard user code, and no automation code.

This is good because it means users won't need to reconfigure their integrations when updating.

### DIFF
--- a/bin/alarm.py
+++ b/bin/alarm.py
@@ -24,7 +24,7 @@ asyncio.set_event_loop(loop)
 panel = Panel(
     host=args.host, port=args.port, 
     automation_code=args.automation_code, 
-    installer_or_user_code=args.installer_code)
+    installer_or_user_code=args.installer_or_user_code)
 try:
     loop.run_until_complete(panel.connect())
     panel.print()
--- a/bin/alarm.py
+++ b/bin/alarm.py
@@ -10,7 +10,7 @@ from bosch_alarm_mode2 import Panel
 cli_parser = argparse.ArgumentParser()
 cli_parser.add_argument('--host', help="panel hostname")
 cli_parser.add_argument('--port', type=int, help="panel port")
-cli_parser.add_argument('-U', '--installer-code', help="Installer code")
+cli_parser.add_argument('-U', '--installer-or-user-code', help="Installer code")
 cli_parser.add_argument('-A', '--automation-code', help="Automation passcode")
 
 args = cli_parser.parse_args()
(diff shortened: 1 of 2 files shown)


bosch_alarm_mode2/panel.py
```diff
@@ -354,9 +356,23 @@ async def _authenticate_automation_user(self):
         raise PermissionError("Authentication failed: " + error)
 
     async def _authenticate(self):
+        if "Solution" in self.model:
```
mag1024, line 359: Let's combine all the logic that acts on the codes in one place: either basicinfo or here.

bin/alarm.py
```diff
@@ -21,7 +21,7 @@
 
 loop = asyncio.new_event_loop()
 asyncio.set_event_loop(loop)
-panel = Panel(host=args.host, port=args.port, automation_code=args.automation_code, installer_code=args.installer_code)
+panel = Panel(host=args.host, port=args.port, automation_code=args.automation_code, installer_or_user_code=args.installer_code)
```
mag1024, line 24: Should probably rename the flag as well?
And line wrap, please

bosch_alarm_mode2/panel.py
```diff
@@ -157,17 +157,10 @@ def __init__(self, host, port, automation_code, installer_code):
...
-                raise ValueError(
-                    "The installer code should only contain numerical digits.")
-            if len(self._installer_code) > 8:
-                raise ValueError(
-                    "The installer code has a maximum length of 8 digits.")
-
     LOAD_BASIC_INFO = 1 << 0
     LOAD_ENTITIES = 1 << 1
     LOAD_STATUS = 1 << 2
+    LOAD_NO_AUTH = 1 << 3
```
mag1024, line 163: Actually... let's rework this slightly.
Entities and status require authentication, so it doesn't make sense for LOAD_NO_AUTH to be part of a mask (mask implies independence).
Also, LOAD_BASIC_INFO now (confusingly) loads extended info -- "basic" is unconditional.

How about we rename LOAD_BASIC_INFO to LOAD_EXTENDED_INFO, and make the call to authenticate conditional to load_selector being non-0?

sanjay900: That sounds reasonable, that was actually what i originally implemented and then wasnt sure what you would prefer

sanjay900: https://github.com/mag1024/bosch-alarm-mode2/pull/29

PR #29, opened 2023-11-01
Load Selector fixes

sanjay900:
Change LOAD_BASIC_INFO to LOAD_EXTENDED_INFO
Remove `LOAD_NO_AUTH` and instead skip auth if the load selector is zero


mag1024:
Thank you!

PR #33, opened 2023-11-22
Add support for getting and setting the panels date and time

sanjay900:
Had a request from bosch for this (plus its something i would find quite useful myself), figure i would just put it behind a service call on the home assistant side


mag1024:
For the integration, I think it might make sense to only provide a function to sync time from HA, not a generalized getter/setter -- I can't imagine how those would be useful in the context of home assistant.

PR #34, opened 2023-11-22
Add support for reading the current panel faults

sanjay900:
Request from bosch, I suspect this would go into another entity so that users can easily see the status of panel faults.

### DIFF
--- a/bosch_alarm_mode2/const.py
+++ b/bosch_alarm_mode2/const.py
@@ -148,19 +148,18 @@ ALARM_MEMORY_PRIORITIES = {
 }
 
 ALARM_PANEL_FAULTS = {
-    0x8000: "Log threshold",
-    0x4000: "Log overflow",
-    0x2000: "Point bus fail since RPS hang up",
-    0x0400: "Fail to call RPS since RPS hang up",
-    0x0200: "User code tamper since RPS hang up",
-    0x0100: "SDI fail since RPS hang up",
-    0x0080: "Communication fail since RPS hang up",
-    0x0020: "AC fail",
-    0x0010: "Battery missing",
-    0x0008: "Battery low",
-    0x0004: "Parameter CRC fail in PIF",
-    0x0002: "Phone line failure"
-    
+    (1 << 1): "Phone line failure",
+    (1 << 2): "Parameter CRC fail in PIF",
+    (1 << 3): "Battery low",
+    (1 << 4): "Battery missing",
+    (1 << 5): "AC fail",
+    (1 << 7): "Communication fail since RPS hang up",
+    (1 << 8): "SDI fail since RPS hang up",
+    (1 << 9): "User code tamper since RPS hang up",
+    (1 << 10): "Fail to call RPS since RPS hang up",
+    (1 << 13): "Point bus fail since RPS hang up",
+    (1 << 14): "Log overflow",
+    (1 << 15): "Log threshold"
 }
 
 class CMD:
--- a/bosch_alarm_mode2/const.py
+++ b/bosch_alarm_mode2/const.py
@@ -148,10 +148,19 @@ ALARM_MEMORY_PRIORITIES = {
 }
 
 ALARM_PANEL_FAULTS = {
-    0x0002: "Phone Line Failure",
-    0x0008: "Battery Low",
-    0x0010: "Battery Missing",
-    0x0020: "AC Failure"
+    0x8000: "Log threshold",
+    0x4000: "Log overflow",
+    0x2000: "Point bus fail since RPS hang up",
+    0x0400: "Fail to call RPS since RPS hang up",
+    0x0200: "User code tamper since RPS hang up",
+    0x0100: "SDI fail since RPS hang up",
+    0x0080: "Communication fail since RPS hang up",
+    0x0020: "AC fail",
+    0x0010: "Battery missing",
+    0x0008: "Battery low",
+    0x0004: "Parameter CRC fail in PIF",
+    0x0002: "Phone line failure"
+    
 }
 
 class CMD:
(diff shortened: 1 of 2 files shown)


bosch_alarm_mode2/const.py
```diff
@@ -147,6 +147,22 @@ class POINT_STATUS:
...
+    0x0400: "Fail to call RPS since RPS hang up",
+    0x0200: "User code tamper since RPS hang up",
+    0x0100: "SDI fail since RPS hang up",
+    0x0080: "Communication fail since RPS hang up",
+    0x0020: "AC fail",
+    0x0010: "Battery missing",
+    0x0008: "Battery low",
+    0x0004: "Parameter CRC fail in PIF",
+    0x0002: "Phone line failure"
+
```
mag1024, line 163: extraneous blank line

bosch_alarm_mode2/const.py
```diff
@@ -147,6 +147,22 @@ class POINT_STATUS:
     0x0A: "Fire Alarm"
 }
 
+ALARM_PANEL_FAULTS = {
+    0x8000: "Log threshold",
```
mag1024, line 151: nit: for all the other maps we list the values in ascending order
Also, since this is a bitmap, might be clear to write the keys (1 << 1), (1 << 2), etc

bosch_alarm_mode2/panel.py
```diff
@@ -142,6 +143,7 @@ def __init__(self, host, port, automation_code, installer_or_user_code):
         self.protocol_version = None
         self.firmware_version = None
         self.serial_number = None
+        self._faults = 0
```
mag1024, line 146: Maybe call this _faults_bitmap, to make it clear that it's not just count?

bosch_alarm_mode2/panel.py
```diff
@@ -436,6 +443,24 @@ async def _extended_info(self):
...
+
+    @property
+    def panel_faults(self) -> [str]:
+        faults = []
+        for mask, fault in ALARM_PANEL_FAULTS.items():
+            if self._faults & mask:
+                faults.append(fault)
+        return faults
+
+    async def _load_panel_status(self):
```
mag1024, line 458: Maybe _load_faults?

bosch_alarm_mode2/panel.py
```diff
@@ -651,6 +680,7 @@ def _on_status_update(self, data):
             0x05: self._area_ready_consumer,
             0x06: self._output_status_consumer,
             0x07: self._point_status_consumer,
+            0x10: self._panel_status_consumer
```
mag1024, line 683: Have you seen these trigger? Don't we also need to subscribe to this type of event?

bosch_alarm_mode2/panel.py
```diff
@@ -436,6 +443,24 @@ async def _extended_info(self):
             revision = int.from_bytes(data[1:2], 'big')
             self.firmware_version = 'v%d.%d' % (version, revision)
 
+    def _set_panel_faults(self, faults):
+        self._faults = faults
+        self.faults_observer._notify()
+
+    @property
+    def panel_faults(self) -> [str]:
+        faults = []
```
mag1024, line 452: Might be able to one line it as a comprehension

bosch_alarm_mode2/panel.py
```diff
@@ -642,6 +667,10 @@ def _event_history_consumer(self, data) -> int:
         self.history_observer._notify()
         return r
 
+    def _panel_status_consumer(self, data) -> int:
+        r = self._set_panel_faults(BE_INT.int16(data,1))
```
mag1024, line 671: _set_panel_faults doesn't return anything

sanjay900:
Trying to figure out what we do if a panel doesn't support set subscriptions command format 2
We could either do it with polling, or we could assume that the panel will probably give us updates as history items, and just do a poll of the panel status when we get a history event?

bosch_alarm_mode2/panel.py
```diff
@@ -436,6 +451,20 @@ async def _extended_info(self):
...
+        self._faults_bitmap = faults
+        self.faults_observer._notify()
+
+    @property
+    def panel_faults(self) -> [str]:
+        return [fault for mask, fault in ALARM_PANEL_FAULTS.items() if self._faults_bitmap & mask]
+
+    async def _load_faults(self):
+        if self._supports_status:
+            data = await self._connection.send_command(
```
mag1024, line 464: nit: one-linify this, and space after ',' on the line below, please

bosch_alarm_mode2/panel.py
```diff
@@ -640,8 +674,16 @@ def _event_summary_consumer(self, data) -> int:
...
         r = self._history.parse_subscription_event(data)
         self.history_observer._notify()
+        # Some panels don't support the subscription for panel status
+        # Since the panel creates history events for most faults
+        # we can just update faults when we get a history event.
+        asyncio.create_task(self._load_faults())
         return r
 
+    def _panel_status_consumer(self, data) -> int:
+        self._set_panel_faults(BE_INT.int16(data,1))
```
mag1024, line 684: Also space after ',' please

bosch_alarm_mode2/panel.py
```diff
@@ -305,6 +318,7 @@ async def _poll(self):
                 await self._load_output_status()
                 await self._get_alarm_status()
```
mag1024, line 319: Looking at this I'm noticing that a) this is very similar to the LOAD_STATUS load mode, and b) the only divergence between the two (the lack of _get_alarm_status in  LOAD_STATUS) is probably not intentional.

Let's make a _load_status helper, and share it from both places.
Also, let's maybe rename _get_alarm_status to _load_alarm_status for consistency with everything else.

mag1024:
Thank you!

PR #35, opened 2023-12-09
Check user privileges before connecting

sanjay900:
A friend of mine was testing the integration, and set up a secondary user specifically for home assistant use. however, they did not give it the `master code functions` authority, which was causing issues as a few commands we use are only accessible to users with that authority. 

This uses command 0x07 (request permission for panel action) to check if the user has access to the `get history` command, which was one of the commands that was noted as not working if you don't have the `master code functions` authority on a solution panel.


bosch_alarm_mode2/panel.py
```diff
@@ -338,6 +338,9 @@ async def _authenticate_remote_user(self):
             await self._connection.send_command(CMD.LOGIN_REMOTE_USER, creds)
         except Exception:
             raise PermissionError("Authentication failed, please check your passcode.")
+        permissions = await self._connection.send_command(CMD.REQUEST_PERMISSION_FOR_PANEL_ACTION, bytearray([AUTHORITY_TYPE.GET_HISTORY]))
```
mag1024, line 341: line wrap, please

bosch_alarm_mode2/const.py
```diff
@@ -162,12 +162,22 @@ class POINT_STATUS:
     (1 << 15): "Log threshold"
 }
 
+class AUTHORITY_TYPE:
+    CHANGE_DATE_TIME = 0x1D
+    CHANGE_USER_PASSCODE = 0x1E
+    ADD_OR_MODIFY_USER = 0x1F
```
mag1024, line 168: Let's only define the value we care about -- GET_HISTORY

bosch_alarm_mode2/panel.py
```diff
@@ -338,6 +338,9 @@ async def _authenticate_remote_user(self):
...
+        if not permissions[0]:
+            raise PermissionError("The specified user does not have the 'master code functions' authority, which is required for this integration.")
```
mag1024, line 343: Maybe just "'master code functions' authority required"
Particularly, "integration" is definitely the right word here.

PR #36, opened 2023-12-19
Fix issues with older AMAX panels

sanjay900:
Seems authentication on the AMAX panels differs once again.

We need to set the user type to installer app, not automation when we authenticate with the automation code.

Also, history ids are worse on the older AMAX panels, and the mask needs to be `0x1FF`.
This does mean we need to mask that out only for AMAX panels, since some of the other panels do allow having more than 512 events.

Also, the AMAX panels don't support the user permission check command, so actually validate that command is available before using it. Not that it matters anyways - when you use the automation code like this, you get all permissions anyways.

### DIFF
--- a/bosch_alarm_mode2/history.py
+++ b/bosch_alarm_mode2/history.py
@@ -25,6 +25,7 @@ class History:
         self._events = []
         self._parser = None
         self._max_count = 0
+        self._amax = False
 
     @property
     def events(self):
@@ -42,6 +43,7 @@ class History:
             self._parser = SolutionHistoryParser()
         elif panel_type <= 0x24:
             self._parser = AmaxHistoryParser()
+            self._amax = True
         else:
             self._parser = BGHistoryParser()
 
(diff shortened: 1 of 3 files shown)


sanjay900:
Ive tested these changes on both fw 1.5 and fw 2.0, I have upgrade keys and a programmer now so I can easily swap between firmwares

bosch_alarm_mode2/history.py
```diff
@@ -42,6 +43,7 @@ def init_for_panel(self, panel_type):
             self._parser = SolutionHistoryParser()
         elif panel_type <= 0x24:
             self._parser = AmaxHistoryParser()
+            self._amax = True
```
mag1024, line 46: Maybe we can make the event_id_mask (or a parse_event_id method) a member of the Parser? This seems like it would make sense, because that's where we presently encapsulate the model-dependent logic.

bosch_alarm_mode2/history.py
```diff
@@ -122,6 +120,9 @@ def parse_subscription_event(self, raw_event) -> HistoryEvent:
     def parse_polled_event(self, id, event_data):
         return HistoryEvent(id, *self._parse_event(self._parse_event_params(event_data)))
 
+    def _parse_event_id(self, event_data):
```
mag1024, line 123: shouldn't have the leading _ 

also maybe it should be called parse_start_event_id, and '+ 1' could stay outside the method -- but up to you

mag1024 (approved):
Good to merge!

PR #38, opened 2024-09-07
Limit the amount of commands in flight at a given time depending on the panel type

sanjay900:
Some bosch panels such as the solution and amax panels get confused if there is more than one packet in flight at a time, so I've added a semaphore to stop this from happening

### DIFF
--- a/bosch_alarm_mode2/connection.py
+++ b/bosch_alarm_mode2/connection.py
@@ -20,7 +20,7 @@ class Connection(asyncio.Protocol):
         self._buffer = bytearray()
         self._pending = deque()
         self._pending_last_empty = datetime.now()
-        self._command_semaphore = asyncio.Semaphore(1)
+        self.set_max_commands_in_flight(1)
 
     def set_max_commands_in_flight(self, command_count):
         self._command_semaphore = asyncio.Semaphore(command_count)
--- a/bosch_alarm_mode2/connection.py
+++ b/bosch_alarm_mode2/connection.py
@@ -18,7 +18,10 @@ class Connection(asyncio.Protocol):
         self._transport = None
         self._buffer = bytearray()
         self._pending = deque()
-        self._transmit_semaphore = asyncio.Semaphore(1)
+        self._command_semaphore = asyncio.Semaphore(1)
+
+    def set_max_commands_in_flight(self, command_count):
+        self._command_semaphore = asyncio.Semaphore(command_count)
 
     def connection_made(self, transport):
         self._transport = transport
@@ -32,9 +35,9 @@ class Connection(asyncio.Protocol):
         self._buffer += data
         self._consume_buffer()
 
-    async def send_command(self, code, data = bytearray()) -> asyncio.Future:
+    async def send_command(self, code, data = bytearray()) -> bytearray:
         # Some panels don't like receiving multiple commands at once, so we limit commands to only run one at a time
-        async with self._transmit_semaphore:
+        async with self._command_semaphore:
             request = bytearray([self.protocol])
             length_size = 2 if self.protocol == PROTOCOL.EXTENDED else 1
             request.extend((len(data) + 1).to_bytes(length_size, 'big'))
--- a/bosch_alarm_mode2/connection.py
+++ b/bosch_alarm_mode2/connection.py
@@ -38,7 +38,8 @@ class Connection(asyncio.Protocol):
         self._consume_buffer()
 
     async def send_command(self, code, data = bytearray()) -> bytearray:
-        # Some panels don't like receiving multiple commands at once, so we limit commands to only run one at a time
+        # Some panels don't like receiving multiple commands at once 
+        # so we limit the amount of commands that are in flight at a given time
         async with self._command_semaphore:
             request = bytearray([self.protocol])
             length_size = 2 if self.protocol == PROTOCOL.EXTENDED else 1
--- a/bosch_alarm_mode2/connection.py
+++ b/bosch_alarm_mode2/connection.py
@@ -18,6 +18,7 @@ class Connection(asyncio.Protocol):
         self._transport = None
         self._buffer = bytearray()
         self._pending = deque()
+        self._transmit_semaphore = asyncio.Semaphore(1)
 
     def connection_made(self, transport):
         self._transport = transport
@@ -31,17 +32,19 @@ class Connection(asyncio.Protocol):
         self._buffer += data
         self._consume_buffer()
 
-    def send_command(self, code, data = bytearray()) -> asyncio.Future:
-        request = bytearray([self.protocol])
-        length_size = 2 if self.protocol == PROTOCOL.EXTENDED else 1
-        request.extend((len(data) + 1).to_bytes(length_size, 'big'))
-        request.append(code)
-        request.extend(data)
-        LOG.debug(">> %s", binascii.hexlify(request))
-        response = asyncio.get_running_loop().create_future()
-        self._pending.append(response)
-        self._transport.write(request)
-        return response
+    async def send_command(self, code, data = bytearray()) -> asyncio.Future:
+        # Some panels don't like receiving multiple commands at once, so we limit commands to only run one at a time
+        async with self._transmit_semaphore:
+            request = bytearray([self.protocol])
+            length_size = 2 if self.protocol == PROTOCOL.EXTENDED else 1
+            request.extend((len(data) + 1).to_bytes(length_size, 'big'))
+            request.append(code)
+            request.extend(data)
+            LOG.debug(">> %s", binascii.hexlify(request))
+            response = asyncio.get_running_loop().create_future()
+            self._pending.append(response)
+            self._transport.write(request)
+            return await response
 
     def close(self):
         if self._transport:
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -464,7 +464,7 @@ class Panel:
         self.model = PANEL_MODEL[data[0]]
         self.protocol_version = 'v%d.%d' % (data[5], data[6])
         # B and G series panels support multiple commands in flight, AMAX and Solution panels do not.
-        if self.model >= 0xA0:
+        if data[0] >= 0xA0:
             self._connection.set_max_commands_in_flight(100)
         if data[13]:
             LOG.warning('busy flag: %d', data[13])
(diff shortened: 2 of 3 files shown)


sanjay900:
This may potentially help with https://github.com/mag1024/bosch-alarm-homeassistant/issues/32 and https://github.com/mag1024/bosch-alarm-homeassistant/issues/31 as well.

sanjay900:
Can confirm this works on the `B5512 (US1B)` as well as the `Solution 4000`. Ill check the AMAX panel and the 3000 as well.

sanjay900:
`Solution 3000` works

sanjay900:
Seems like the priorities do read correctly from the AMAX panel too, had to remember how to update and configure it

mag1024:
I did some testing, and I think I was able to reproduce the problem.
From what I can tell, it doesn't have anything to do with concurrent memory commands per se -- rather, I believe, it is the `output status` command that is busted, and sometimes fails to return any response, causing havoc on response ordering.
Could you please try disabling `_load_output_status`, and seeing if it fixes this problem on Solution panels?

sanjay900:
Interesting, I'll give that a go when I get home.

sanjay900:
> I did some testing, and I think I was able to reproduce the problem. From what I can tell, it doesn't have anything to do with concurrent memory commands per se -- rather, I believe, it is the `output status` command that is busted, and sometimes fails to return any response, causing havoc on response ordering. Could you please try disabling `_load_output_status`, and seeing if it fixes this problem on Solution panels?

Since I don't have outputs configured on my panel, it would have been exiting that function early, so all my testing was done without that command being involved.

mag1024:
I'm basically trying to confirm the theory that this is somehow related to concurrent memory commands.
To do that, I've been running a function like this several times after connect:

```
async def memory_test(self):
    await asyncio.gather(
        self._load_faults(),
        self._load_history(),
        self._load_faults(),
        self._load_history(),
        self._load_faults(),
        self._load_history(),
        self._load_faults(),
        self._load_history(),
        self._get_alarms_for_priority(0x07),
        self._get_alarms_for_priority(0x09),
        self._get_alarms_for_priority(0x0A)
    )
```

I'm also logging the request for each response we process, so I can see if they ever get mismatched.
So far, the only time I've seen it fail -- and for a while I had it failing pretty consistently -- has been when there's also an output status query happening, and it is always the one that drops the response, not any of the memory commands. So it definitely seems buggy, but maybe some other commands are too.

sanjay900:
One thing to note is the panels get worse at responding when they are actually armed as well. I know for a fact that the memory related commands especially get a bit whack - commands for history for example just stop working all together when you arm the panel.

mag1024:
I did more testing, including while armed, and over 100+ runs, it hasn't failed once without output status fetching -- whereas with it enabled, it seem to fail somewhat regularly.

I pushed the hacky code I'm using for testing to memory_test branch. Could you please give it a shot with your panels?
You can run it in a shell for-loop, because if there is a missing response, the last command will get stuck waiting for a response.

sanjay900:
What panel are you testing this with?

sanjay900:
It was happening whenever I set the alarm off, and it happened every time on my 4000 before I added that semaphore.

I'll see if that happens with my 3000 as well and run your stress test.

mag1024:
I'm testing with my old faithful B8512G.
The reason I'm trying to confirm if there's actually an issue with concurrent memory commands, is that I suspect that while the semaphore happens to help ameliorate the triggering issue, it does so indirectly, and as such will not be reliable.
If the issue is not specific to the memory commands, but rather with the panel being "too loaded" or some such while triggering, and randomly dropping commands, this is probably not the right approach to fixing it.

EHylands:
Hi @sanjay900 and @mag1024 
Some words for thought.

Had same out of sync and corrupted responses with outputs and alarms details function on Solution panels. 

Mostly appeared when protocol 0x01 function (get outputs states and get alarm details) gets called from async protocol 0x02 before previous command answer has been received. I used Sanjay semaphore solution to fix the issue (1 semaphore shared amongst all protocol 0x01 calls).

Did extensive documentation reading and no information is given regarding commands concurrency nor about any guaranteed response order of arrival.

mag1024:
> Did extensive documentation reading and no information is given regarding commands concurrency nor about any guaranteed response order of arrival.

For a streaming protocol like this with no message ids, the _only_ sane behavior is to guarantee order, and responses. If ever the panel needs to drop some messages, it _must_ NACK, or reset the connection. Dropping and/or reordering responses is always a bug in the firmware.

> Mostly appeared when protocol 0x01 function (get outputs states and get alarm details) gets called from async protocol 0x02 before previous command answer has been received. 

Have you tried disabling output polling? AFAICT, their implementation in the firmware is very half-baked. Sanjay already ran into concurrency issues with setting of the outputs, the output ids used in subscription events are unusable, and now it looks like the output get state is prone to getting dropped.

sanjay900:
> > Did extensive documentation reading and no information is given regarding commands concurrency nor about any guaranteed response order of arrival.
> 
> For a streaming protocol like this with no message ids, the _only_ sane behavior is to guarantee order, and responses. If ever the panel needs to drop some messages, it _must_ NACK, or reset the connection. Dropping and/or reordering responses is always a bug in the firmware.
> 
> > Mostly appeared when protocol 0x01 function (get outputs states and get alarm details) gets called from async protocol 0x02 before previous command answer has been received.
> 
> Have you tried disabling output polling? AFAICT, their implementation in the firmware is very half-baked. Sanjay already ran into concurrency issues with setting of the outputs, the output ids used in subscription events are unusable, and now it looks like the output get state is prone to getting dropped.

The thing is, it can't just be the output polling since i was running into this problem even without output polling being enabled. It sounds more like we just have to throw a semaphore on all commands, so that we don't ever get in the position where more than one command is on the wire at a time.

sanjay900:
> I'm testing with my old faithful B8512G. The reason I'm trying to confirm if there's actually an issue with concurrent memory commands, is that I suspect that while the semaphore happens to help ameliorate the triggering issue, it does so indirectly, and as such will not be reliable. If the issue is not specific to the memory commands, but rather with the panel being "too loaded" or some such while triggering, and randomly dropping commands, this is probably not the right approach to fixing it.

Just using your branch to test, if i run any two commands at the same time with gather, the first one gets a response and the second one doesn't. For me, it just hangs like its waiting for a response that never arrives if i do that.

sanjay900:
Oh my god i think this might be what was causing the network module to hang.

sanjay900:
Seems that it is only the solution and amax panels that just can't tolerate multiple commands simultaneously like that, the B5512 I have does not seem to care.

bosch_alarm_mode2/connection.py
```diff
@@ -31,17 +32,19 @@ def data_received(self, data):
...
-        length_size = 2 if self.protocol == PROTOCOL.EXTENDED else 1
-        request.extend((len(data) + 1).to_bytes(length_size, 'big'))
-        request.append(code)
-        request.extend(data)
-        LOG.debug(">> %s", binascii.hexlify(request))
-        response = asyncio.get_running_loop().create_future()
-        self._pending.append(response)
-        self._transport.write(request)
-        return response
+    async def send_command(self, code, data = bytearray()) -> asyncio.Future:
```
mag1024, line 35: No longer accurate

bosch_alarm_mode2/connection.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, passcode, on_status_update, on_disconnect):
         self._transport = None
         self._buffer = bytearray()
         self._pending = deque()
+        self._transmit_semaphore = asyncio.Semaphore(1)
```
mag1024, line 21: since the critical section includes waiting for the response, not just transmitting the command, `_command_semaphore` might be more appropriate

mag1024, line 21: It's really unfortunate to limit all panels to a single in flight command because some older panels can't deal with it. Let's set the limit based on the model. For more capable panels, I think we can set a high limit here, and deal with the triggering issue with a combination of #39, and converting the subscription handlers to schedule followups with a delay.

mag1024 (changes requested):
Cool, I like this a lot more, because this feels much more principled than randomly mutexing _some_ commands.
Please update the PR description.

mag1024:
@sanjay900, should we also report the issues with response dropping as bugs to Bosch?
The output implementation in particular seems quite busted -- and it seems that other commands can also be problematic when the panel is loaded, such as with triggering.

bosch_alarm_mode2/connection.py
sanjay900: Yeah I guess we just have to decide what a reasonable limit is for the b and g panels

sanjay900:
> @sanjay900, should we also report the issues with response dropping as bugs to Bosch? The output implementation in particular seems quite busted -- and it seems that other commands can also be problematic when the panel is loaded, such as with triggering.

I've done so already and given them a link to this thread as well

bosch_alarm_mode2/connection.py
mag1024: Up until now, it has been unlimited so... 10? 100?

bosch_alarm_mode2/panel.py
```diff
@@ -464,6 +463,9 @@ async def _basicinfo(self):
             data = await self._connection.send_command(CMD.WHAT_ARE_YOU)
         self.model = PANEL_MODEL[data[0]]
         self.protocol_version = 'v%d.%d' % (data[5], data[6])
+        # B and G series panels support multiple commands in flight, AMAX and Solution panels do not.
+        if self.model >= 0xA0:
```
mag1024, line 467: self.model is a string model descriptor -- this should be `data[0]`

bosch_alarm_mode2/connection.py
```diff
@@ -20,6 +20,10 @@ def __init__(self, passcode, on_status_update, on_disconnect):
         self._buffer = bytearray()
         self._pending = deque()
         self._pending_last_empty = datetime.now()
+        self._command_semaphore = asyncio.Semaphore(1)
```
mag1024, line 23: can do `self.set_max_commands_in_flight(1)` instead

mag1024 (approved):

sanjay900:
@mag1024 Had someone reach out from bosch, sounds like its by design that the protocol only works with one command at a time. Seems like they sort of expect you to use the multiple encapsulation protocol if you want to send more than one command at a time.

mag1024:
Well, that's a pretty underwhelming response. Clearly, as indicated by the fact that this work fine of with B/G panels, this is a limitation of the _implementation_, not the the _design_ of the protocol.

Implementing an incoming command queue, up to a reasonable buffer limit, is straightforward. However, even if Bosch is unwilling to do that on their low end panels, there is, in my opinion, no excuse for the failure to NACK dropped command, or, at the barest minimum, clearly documenting the expectations in the protocol spec.

PR #39, opened 2024-09-10
Reset the connection if possible command skew due to dropped responses is detected

mag1024:


sanjay900:
Looks reasonable to me

PR #41, opened 2024-09-10
Allow users without history access to authenticate

### LINKED ISSUE #40: Permission check causes problems if a panel is currently armed
A while back i implemented a permission check, with https://github.com/mag1024/bosch-alarm-mode2/pull/35
At the time i validated privileges by checking for history privileges. This however causes issues if a panel is armed, as the panels lock access to the history memory while the panel is armed. This means a user is not able to connect to an armed panel.

We will have to find some other way to validate credentials.
I wonder if any of the other permissions we can check for are enough to validate if a user has master code functions or not.

sanjay900:
Closes #40 

We previously did a permission check to handle the case where a user doesn't have permission to get history events, but we can just do a try catch instead to handle the cases where a user can't run that command

This fixes an issue where you can't actually authenticate with a panel right now if it is armed, as a panel that is armed revokes permissions for reading history.

This means that now the integration is usable even if you don't have master code authority over a panel, it just stops history from working in that scenario.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -320,28 +320,22 @@ class Panel:
             await self._load_entity_status(CMD.DOOR_STATUS, self.doors, 1)
 
     async def _load_history(self):
-        # Don't retrieve history when in any state that isn't disarmed, as panels do not support this.
-        if not all(area.is_disarmed() for area in self.areas.values()):
-            return
-        # Check if the current user has permission to read the event history
-        if self._supports_permission_check:
-            permissions = await self._connection.send_command(
-                CMD.REQUEST_PERMISSION_FOR_PANEL_ACTION, bytearray([AUTHORITY_TYPE.GET_HISTORY]))
-            if not permissions[0]:
-                return
-        start_size = len(self.events)
-        start_t = time.perf_counter()
-        event_id = self._history.last_event_id
-        while event_id is not None:
-            request = bytearray(b'\xFF')
-            request.extend(event_id.to_bytes(4, 'big'))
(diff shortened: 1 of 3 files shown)


sanjay900:
I'm just not sure how I feel about doing an extra command every time we poll history but I guess for most panels with subscriptions that doesn't matter

sanjay900:
Is there a neater way to deal with the exception, being that we don't have a specific exception type for the nacks or anything like that?

sanjay900:
I guess the only problem with doing it like this is that we would flood logs on panels that don't support subscriptions.

bosch_alarm_mode2/panel.py
```diff
@@ -320,22 +319,22 @@ async def _load_status(self):
...
+                request.extend(event_id.to_bytes(4, 'big'))
+                data = await self._connection.send_command(self._history_cmd, request)
+                self._last_msg = datetime.now()
+                if (event_id := self._history.parse_polled_events(data)):
+                    self.history_observer._notify()
+            if len(self.events) != start_size:
+                LOG.debug("Loaded %d history events in %.2fs" % (
+                    len(self.events) - start_size, time.perf_counter() - start_t))
+        except Exception:
+            LOG.exception("Failed to load history events")
```
mag1024, line 337: Let's do something like

LOG.warning("Failed to load history events; ensure panel is not armed, and user has 'master code functions' authority.")

and set some bool inside _history so that we don't keep spamming the logs.

sanjay900: ill bring back the check for if the panel is armed then, that way a panel being armed won't set this flag

PR #42, opened 2024-09-17
Reduce number of followup commands for area on/off events.

mag1024:
Hopefully this will also reduce the probability of triggering dropped commands.

1) Only fetch the history if it is empty.
This is not 100% correct, because it is possible that though the initial load didn't happen because the panel was armed, some history were delivered via the subscription. However, it seems we already don't handle that, because the inserted history event will cause last_event_id to be current.

2) Drop load_faults on area on/off.
Thinking about this more, it is not actually necessary. Even if the history could not be fetched at startup, the faults should've been. If there were any history events since, those would already have been followed up.

3) Delay the history fetch.
Sleeps are gross, but it does seem that the panels are somehow more prone to losing commands immediately after disarm.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -122,7 +122,7 @@ class Door(PanelEntity):
 
     def __repr__(self):
         return f"{self.name}: {DOOR_STATUS.TEXT[self.status]}"
-    
+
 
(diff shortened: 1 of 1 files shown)


sanjay900 (approved):
Looks reasonable to me

PR #45, opened 2024-12-26
Add 'connection established' log info msg to connection.py

david9600:
Add an info log message to indicate successful connection to the panel. This would be useful when investigating a config or network issue, since it gives immediate feedback, particularly now that the HA system log auto-updates.

### DIFF
--- a/bosch_alarm_mode2/connection.py
+++ b/bosch_alarm_mode2/connection.py
@@ -26,7 +26,7 @@ class Connection(asyncio.Protocol):
         self._command_semaphore = asyncio.Semaphore(command_count)
 
     def connection_made(self, transport):
-        LOG.info("Panel connected.")
+        LOG.info("Connection established.")
         self._transport = transport
 
     def connection_lost(self, exc):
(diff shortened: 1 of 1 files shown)


bosch_alarm_mode2/connection.py
```diff
@@ -26,6 +26,7 @@ def set_max_commands_in_flight(self, command_count):
         self._command_semaphore = asyncio.Semaphore(command_count)
 
     def connection_made(self, transport):
+        LOG.info("Panel connected.")
```
mag1024, line 29: By analogy with line 32: "Connection established."

david9600: Fine by me ... do I need to edit the PR? (which I'm not sure how to do, other than submitting a new one!)

sanjay900: Push another commit to your branch, and it will then update the PR

david9600: OK... I think I've managed to do that (using GitHub web interface).

PR #48, opened 2025-02-23
Tooling and project metadata updates

mag1024:
- Switch from poetry to uv
- Specify the license as MIT
- Use ruff, and apply its formatting and lint fixes


sanjay900 (approved):
Looks good to me

sanjay900:
Does swapping to uv make it easier to set up ci?

mag1024:
The CI example I was looking at uses uv, and that's what HA itself switched to, so I figure it's time.

PR #50, opened 2025-03-01
Add typing

sanjay900:
Figured since we had a lot of types specified it would be worth it to specify them everywhere.

Worked out to be a useful spend of time as it actually did find a typo.

With this, i can specify that the home assistant integration is staticly typed and it passes mypy, which is one of the requirements if we ever want to get past being a bronze integration and get all the way to platnium


bosch_alarm_mode2/connection.py
```diff
@@ -1,45 +1,47 @@
...
 
-    def data_received(self, data):
+    def data_received(self, data: bytes) -> None:
         LOG.debug("<< %s", binascii.hexlify(data))
         self._buffer += data
         self._consume_buffer()
 
-    async def send_command(self, code, data=bytearray()) -> bytearray:
+    async def send_command(self, code: int, data: bytes=bytearray()) -> bytearray | None:
+        if not self._transport:
```
mag1024, line 43: Hm, maybe throw here instead?

bosch_alarm_mode2/panel.py
```diff
@@ -1,10 +1,34 @@
...
 from datetime import datetime, timedelta
-
-from .const import *
+from typing import Any, Generator
+
+from .const import (
+    ALARM_MEMORY_PRIORITIES,
+    ALARM_MEMORY_PRIORITY_ALARMS,
+    ALARM_PANEL_FAULTS,
+    AREA_ARMING_AWAY,
```
mag1024, line 13: Let's convert some of these loose consts into classes, like we're doing with, for example, POINT_STATUS to make this import a little less verbose.

bosch_alarm_mode2/panel.py
```diff
@@ -17,94 +41,94 @@
...
         return self.status in AREA_STATUS.ALL_ARMED
 
-    def is_armed(self):
+    def is_armed(self) -> bool:
         return self.status in AREA_STATUS.ARMED
 
-    def is_triggered(self):
-        return (self.is_armed() or self.is_pending()) and self._alarms.intersection(
+    def is_triggered(self) -> bool:
+        return bool((self.is_armed() or self.is_pending()) and self._alarms.intersection(
```
mag1024, line 122: Why is this necessary?

bosch_alarm_mode2/panel.py
```diff
@@ -353,14 +378,16 @@ async def _load_status(self):
         if self._supports_door:
             await self._load_entity_status(CMD.DOOR_STATUS, self.doors, 1)
 
-    async def _load_history(self):
+    async def _load_history(self) -> None:
+        if not self._connection or not self._history_cmd:
```
mag1024, line 382: For _connection, I think it would be more idiomatic to throw from send_command on error, and not have to check it everywhere.

And for _history_cmd -- is there a way we can arrive here without having initialized it?
If there is, a better solution might be to default _history_cmd to REQUEST_RAW_HISTORY_EVENTS in the constructor, and to "upgrade" it in basicinfo

bosch_alarm_mode2/panel.py
```diff
@@ -413,7 +440,8 @@ async def _monitor_connection_once(self):
 
         idle_time = datetime.now() - (self._last_msg or datetime.fromtimestamp(0))
         if idle_time > timedelta(minutes=3):
-            LOG.warning("Heartbeat expired (%s): resetting connection.", idle_time)
+            LOG.warning(
```
mag1024, line 443: Is this how ruff formats?
https://github.com/mag1024/bosch-alarm-mode2/blob/af3bd3f655f5587ab6ba2c6f749856b40e28e998/Makefile#L15

pyproject.toml
```diff
@@ -41,3 +41,6 @@ line-length = 100
 
 [tool.ruff.lint]
 ignore = ["F403", "F405", "E722"]
+
+[tool.setuptools.package-data]
+"bosch-alarm-mode2" = ["py.typed"]
```
mag1024, line 46: Does this do something different than creating a py.typed file?
https://docs.astral.sh/uv/concepts/projects/init/#libraries

mag1024 (changes requested):

bosch_alarm_mode2/panel.py
sanjay900: `_alarms.intersection` returns the intersection of the sets as a set, you'd expect is_triggered to return a bool so we need to cast that result to bool

bosch_alarm_mode2/connection.py
sanjay900: it would still be possible to return None, as `_process_response` sets the result to `None` after a `0xFC`
We could just change that to return an empty byte array though and then if we combine that with throwing we don't have to check data everywhere?

bosch_alarm_mode2/panel.py
sanjay900: youd still have to check _connection everywhere since send_command lives in _connection

bosch_alarm_mode2/panel.py
mag1024: Ah, my C++-honed intuition misleads me.
Still I think it would be clearer to cast only the result of self._alarms.intersection(...) to bool -- the other too components should be bools already

Same for connection_status()

bosch_alarm_mode2/panel.py
mag1024: Then let's make a _send_command helper on the Panel that checks the connection, and possibly throws.

Not only would this be less verbose, it would, I think, also be more correct: some of these functions execute multiple commands and the state of `_connection` can change

bosch_alarm_mode2/connection.py
mag1024: Yep, I like that idea!

bosch_alarm_mode2/panel.py
```diff
@@ -451,7 +475,9 @@ async def _authenticate_automation_user(self, user_type):
         error = ["Not Authorized", "Authorized", "Max Connections"][result[0] if result else 0]
         raise PermissionError("Authentication failed: " + error)
 
-    async def _authenticate(self):
+    async def _authenticate(self) -> None:
+        if not self.model:
+            return
```
mag1024, line 480: I see that the strict typing has indeed uncovered some interesting corner case we're not handling, but I don't think that silently swallowing the errors is the right way to fix them.

Once we fix `send_command` to never return None, I believe `basicinfo` will always populate model or throw, so it should be impossible to have an empty model here. Can we hint this to the type checker somehow?

bosch_alarm_mode2/panel.py
sanjay900: Cool, easy I'll do that

bosch_alarm_mode2/connection.py
```diff
@@ -12,34 +13,37 @@
...
-    def data_received(self, data):
+    def data_received(self, data: bytes) -> None:
         LOG.debug("<< %s", binascii.hexlify(data))
         self._buffer += data
         self._consume_buffer()
 
-    async def send_command(self, code, data=bytearray()) -> bytearray:
+    async def send_command(self, code: int, data: bytes = bytearray()) -> bytearray:
+        if not self._transport:
+            raise asyncio.InvalidStateError()
```
mag1024, line 46: "transport not connected"

bosch_alarm_mode2/const.py
```diff
@@ -99,22 +99,25 @@ class OUTPUT_STATUS:
...
+
 
-AREA_ARMING_DISARM = 0x01
-AREA_ARMING_MASTER_DELAY = 0x03
-AREA_ARMING_PERIMETER_DELAY = 0x05
-AREA_ARMING_STAY1 = 0x0A
-AREA_ARMING_STAY2 = 0x0B
-AREA_ARMING_AWAY = 0x0C
+class AREA_ARMING_STATUS:
+    AREA_ARMING_DISARM = 0x01
```
mag1024, line 115: Can probably drop the "AREA_ARMING" prefix on all the values -- same for AREA_READY above

bosch_alarm_mode2/panel.py
```diff
@@ -331,7 +350,12 @@ def connection_factory():
             await self.load(load_selector)
         self.connection_status_observer._notify()
 
-    def _on_disconnect(self):
+    async def _send_command(self, code: int, data: bytes = bytearray()) -> bytearray:
+        if not self._connection:
+            raise asyncio.InvalidStateError()
+        return await self._connection.send_command(code, data)
```
mag1024, line 356: is the "await" here necessary?

bosch_alarm_mode2/panel.py
```diff
@@ -422,36 +446,35 @@ async def _monitor_connection_once(self):
...
             creds = int(str(self._installer_or_user_code).ljust(8, "F"), 16)
-            creds = creds.to_bytes(4, "big")
-            await self._connection.send_command(CMD.LOGIN_REMOTE_USER, creds)
+            await self._send_command(CMD.LOGIN_REMOTE_USER, creds.to_bytes(4, "big"))
         except Exception:
             raise PermissionError("Authentication failed, please check your passcode.")
 
-    async def _authenticate_automation_user(self, user_type):
+    async def _authenticate_automation_user(self, user_type: int) -> None:
+        if not self._automation_code:
```
mag1024, line 464: The only call to this function is gated by `if self._automation_code:`

mag1024 (approved):

bosch_alarm_mode2/panel.py
mag1024: Also, `InvalidStateError("not connected")`

bosch_alarm_mode2/panel.py
sanjay900: yeah i might just have to tell it to skip typing errors on that line i dont know if theres a better way around the type checking there

bosch_alarm_mode2/panel.py
sanjay900: I suspect the await is necessary because the function is async
Could make it a not async function that returns an awaitable byte array though otherwise?

sanjay900: but that just seems less clean

PR #54, opened 2025-04-05
Fix AMAX history event id parsing

sanjay900:
After playing with my AMAX panel, ive come to the conclusion that it stops recording log entries when all 256 entries are filled up, and in that scenario the event id just keeps rolling over and we read events forever. 

The event ids seem to start at 256 and end at 512.

To avoid this, i just stop reading events if we get an event id > 512.

### DIFF
--- a/bosch_alarm_mode2/history.py
+++ b/bosch_alarm_mode2/history.py
@@ -226,6 +226,8 @@ class History:
         for i in range(start, start + count):
             try:
                 e = self._parser.parse_polled_event(i, event_data)
+                if self._events and e.date < self._events[-1].date:
+                    return None
                 LOG.debug(e)
                 self._events.append(e)
                 event_data = event_data[event_length:]


bosch_alarm_mode2/history.py
```diff
@@ -16,6 +16,7 @@
 
 class HistoryEvent(NamedTuple):
     id: int
+    raw_id: int
```
mag1024, line 19: So what happens when we use the masked id for polling with Amax, like we were doing previously?

After this change, if the top bits are event type, doesn't it mean that once you get a "higher" event type, subsequent polls will not retrieve lower types?

sanjay900: Honestly I haven't really figured it out but when I did anything else but this by panel would just get stuck in an infinite loop of polling

sanjay900:
Honestly part of me is just tempted to disable history for AMAX panels otherwise, unless i get more info about exactly how its supposed to work from bosch.

sanjay900:
I think what happens with these panels is the event ids just roll over internally, and so there isnt a "last event id" as such.

I don't really think there is a solution to this.

mag1024:
Oh, wow, it even says in the amax [brochure](https://cdn.commerce.boschsecurity.com/public/documents/AMAX_panel_4000_Data_sheet_enUS_22657011723.pdf) that it only supports 256 events -- but apparently of 3 different types?

> The only real solution is to read all 512 events from the panel, and then when reading events we would need to stop if we have already seen the event

What if we kept all the logic exactly as is, but didn't push the event (and consequently didn't update the last_event_id), if the timestamp of the new event is < the last recorded event?

sanjay900:
> Oh, wow, it even says in the amax [brochure](https://cdn.commerce.boschsecurity.com/public/documents/AMAX_panel_4000_Data_sheet_enUS_22657011723.pdf) that it only supports 256 events -- but apparently of 3 different types?
> 
> > The only real solution is to read all 512 events from the panel, and then when reading events we would need to stop if we have already seen the event
> 
> What if we kept all the logic exactly as is, but didn't push the event (and consequently didn't update the last_event_id), if the timestamp of the new event is < the last recorded event?

that works for me let me give it a go

sanjay900:
That works perfectly

PR #56, opened 2025-04-06
Allow getting faults by ids so we can build nicer sensors in HA

sanjay900:
I was talking to one of the devs and it was pointed out that things like faults would be better as separate binary sensors instead of just a single sensor with text in it, as that would allow people to do automation for specific faults easier.

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -116,7 +116,7 @@ class Area(PanelEntity):
 
     def is_triggered(self) -> bool:
         return (self.is_armed() or self.is_pending()) and bool(
-            self._alarms.intersection(ALARM_MEMORY_PRIORITIES.PRIORITY)
+            self._alarms.intersection(ALARM_MEMORY_PRIORITIES.PRIORITY_ALARMS)
         )
 
     def reset(self) -> None:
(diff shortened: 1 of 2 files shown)


bosch_alarm_mode2/panel.py
```diff
@@ -79,7 +78,11 @@ def faults(self) -> int:
 
     @property
     def alarms(self) -> list[str]:
-        return [ALARM_MEMORY_PRIORITIES[x] for x in self._alarms]
+        return [ALARM_MEMORY_PRIORITIES.TEXT[x] for x in self._alarms]
+
+    @property
+    def alarms_ids(self) -> set[int]:
```
mag1024, line 84: interface-wise, it's kind of odd that everything else is a list, and this is a set

mag1024 (approved):
This change is fine, but I'm not really convinced that separate sensors for everything here is a good idea -- especially for faults. Many of them seem quite esoteric, and not likely to be useful in automations. 
Being presented with a very long list of unuseful entities is not a good user experience.

For the alarms -- this does seem potentially useful for automations, but I wouldn't represent this is a whole bunch of binary_sensors. It seems that there are really 3 alarm types (burglary, gas, fire), each with 3 levels (trouble, supervisory, and alarm). I would go with a 4 state sensor for each. Maybe I'll comment on the HA PR.

bosch_alarm_mode2/panel.py
sanjay900: yeah thats fair, ill change it to a list

PR #60, opened 2025-11-03
Enable arm_part_instant (Night Mode) for BG Panels

EHylands:
### Summary
This PR enables Night Mode (arm_part_instant) support for BG Panels.
Currently, when arming the perimeter, a delay is applied before triggering an alarm if a point is opened. This update allows instant triggering at night when a perimeter point trips, consistent with “Perimeter Instant Arm” behavior on Bosch B/G-Series panels.

This functionality could also align with the Stay2 arming mode on Bosch Solution panels.
Night Mode can later be exposed automaticaly or as an option in the Home Assistant integration.

### Changes
**const.py**
Added:
class AREA_ARMING_STATUS:
    PERIMETER_INSTANT = 0x04

**panel.py**
Added _partial_arming_instant_id variable
Added area_arm_part_instant() function
Added is_arm_part_instant_supported() function

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -278,22 +278,10 @@ class Panel:
         await self._area_arm(area_id, AREA_ARMING_STATUS.DISARM)
 
     async def area_arm_part(self, area_id: int, delay: bool = True) -> None:
-        part_delay_id, part_instant_id = self._partial_arming_id
-
-        part_arming_id = part_delay_id
-        if not delay and part_instant_id is not None:
-            part_arming_id = part_instant_id
-        
-        await self._area_arm(area_id, part_arming_id)
+        await self._area_arm(area_id, self._get_arming_id(delay, *self._partial_arming_id))
 
     async def area_arm_all(self, area_id: int, delay: bool = True) -> None:
-        all_delay_id, all_instant_id = self._partial_arming_id
-
-        all_arming_id = all_delay_id
-        if not delay and all_instant_id is not None:
-            all_arming_id = all_instant_id
-        
-        await self._area_arm(area_id, all_arming_id)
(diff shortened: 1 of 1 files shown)


bosch_alarm_mode2/const.py
```diff
@@ -114,6 +114,7 @@ class AREA_READY_STATUS:
 class AREA_ARMING_STATUS:
     DISARM = 0x01
     MASTER_DELAY = 0x03
+    PERIMETER_INSTANT = 0x04
```
mag1024: Let's also add the master instant while we're here.

mag1024 (changes requested):
Thanks for the PR! The feature makes sense -- some comments in the direction of making more general.

bosch_alarm_mode2/panel.py
```diff
@@ -530,11 +542,11 @@ async def _basicinfo(self) -> None:
...
         # Solution and AMAX panels use different arming types from B/G series panels.
         if data[0] <= 0x28:
-            self._partial_arming_id = AREA_ARMING_STATUS.STAY1
-            self._all_arming_id = AREA_ARMING_STATUS.AWAY
+            self._partial_arming_id = (AREA_ARMING_STATUS.STAY1, None)
+            self._all_arming_id = (AREA_ARMING_STATUS.AWAY, None)
         else:
-            self._partial_arming_id = AREA_ARMING_STATUS.PERIMETER_DELAY
-            self._all_arming_id = AREA_ARMING_STATUS.MASTER_DELAY
+            self._partial_arming_id = (AREA_ARMING_STATUS.PERIMETER_DELAY, AREA_ARMING_STATUS.PERIMETER_INSTANT)
```
mag1024: Probably don't need this else at all, given the defaults in the constructor

PR #61, opened 2025-11-09
part_armed_enabled

EHylands:
- Added `PART_ARMED_INSTANT` and  `PART_ARMED_DELAY` in AREA_STATUS constant
- Added `BG_PANELS` constant
- Implemented or updated the following methods:
  - `is_part_armed_delay(self) ...`
  - `is_part_armed_instant(self) ...`
  - `is_part_armed(self) ...`
  - `is_part_arm_instant_enabled(self) ...`

Only BG panels can:
- Arm an area with `PART_ON_INSTANT (0x02)`
- Report an area `PART_ON_INSTANT (0x02)` status

### DIFF
--- a/bosch_alarm_mode2/panel.py
+++ b/bosch_alarm_mode2/panel.py
@@ -126,7 +126,7 @@ class Area(PanelEntity):
         return (self.is_armed() or self.is_pending()) and bool(
             self._alarms.intersection(ALARM_MEMORY_PRIORITIES.PRIORITY_ALARMS)
         )
-    
+
     def reset(self) -> None:
         self.status = AREA_STATUS.UNKNOWN
         self._set_ready(AREA_READY_STATUS.NOT, 0)


EHylands:
Request Area Authorities [0x25] with 0x0A can check for perimeter instant ARM authority. In documentation, it states it will always return true for automation user. Would have to check result on Solution Panl !

bosch_alarm_mode2/const.py
```diff
@@ -62,13 +62,26 @@
     0xA9: "B6512 (US1B)",
 }
 
+BG_PANELS = {
```
mag1024: We should have one place where we list the panel models.
Let's have an enum to indicate the family, and have PANEL_MODEL map to a named tuple that includes the model name and the family.

bosch_alarm_mode2/const.py
```diff
@@ -45,21 +48,32 @@
...
-    0x24: "AMAX 4000",
-    0x79: "D7412GV4",
-    0x84: "D9412GV4",
-    0xA0: "B4512 (US1B)",
-    0xA4: "B5512 (US1B)",
-    0xA6: "B8512G (US1A)",
-    0xA7: "B9512G (US1A)",
-    0xA8: "B3512 (US1B)",
-    0xA9: "B6512 (US1B)",
+class PANEL_FAMILY():
```
mag1024: Enum

bosch_alarm_mode2/const.py
```diff
@@ -45,21 +48,32 @@
...
+    BG_SERIE = "BG_SERIE"
+    SOLUTION = "SOLUTION"
+    AMAX = "AMAX"
+
+
+class PANEL_INFO(NamedTuple):
```
mag1024: @dataclass(frozen=True) class PanelModel:
  name: str
  family: "PanelFamily"

(notice how dropping the model_ prefix will improve the readability of use in panel.py)

bosch_alarm_mode2/panel.py
```diff
@@ -118,7 +126,7 @@ def is_triggered(self) -> bool:
         return (self.is_armed() or self.is_pending()) and bool(
             self._alarms.intersection(ALARM_MEMORY_PRIORITIES.PRIORITY_ALARMS)
         )
-
+
```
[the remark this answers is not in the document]
mag1024: Please fix

PR #69, opened 2026-09-20
Delay fault check after history events to avoid Solution panel hang on arm

### LINKED ISSUE #68: Solution 3000 stops responding on arm/alarm when fault check (0x20) is sent immediately after history event
**Panel:** Solution 3000, firmware 2.1.3 · **IP module:** B426-M 03.16.202 · **Library:** 0.4.10 (via HA `bosch_alarm`)

**Symptom:** All entities go `unavailable` for ~40 s on most arms and during alarms. Log shows `Detected possible command skew: resetting connection.`

**Cause (from debug logs):** When the exit delay ends, the panel pushes the "AWAY/STAY Arm" history event. `_event_history_finalizer` immediately schedules `_load_faults()`, so `REQUEST_PANEL_SYSTEM_STATUS (0x20)` is sent within ~1 ms. In the failing cases the panel then goes completely silent on the automation session: it never answers 0x20, and it also stops sending the area-status push and even its 60 s heartbeat. Because Solution panels allow only one command in flight, the watchdog's `WHAT_ARE_YOU` probe queues behind the stuck 0x20, times out, and the connection is reset.

Failing example:
```
14:52:38.032 << 02002b0201...  [694] Quick Area0 AWAY Arm
14:52:38.033 >> 010120
  (nothing from the panel, not even the heartbeat due ~14:53:11)
14:52:42.046 Checking for command skew (0:01:03)...
14:53:12.046 Detected possible command skew: resetting connection.
```
Successful example (same panel), with the area push arriving before the 0x20 reply:
```
15:36:04.002 << ... [700] Quick Area0 AWAY Arm
15:36:04.003 >> 010120
15:36:04.218 << 0200050401000101   Area 1: All On / Away Armed
15:36:04.363 << 010efe02010321...
```

**Results:**
- As-is: **4 of 5 arms** (stay and away) reset the connection; 6/6 disarms fine. Earlier captures show the same 0x20 hang on alarm history events while armed.
- Delaying the fault check by 5 s after a history event: **5 of 5 arms + 5 of 5 disarms clean**. The panel sends its area push normally, then answers 0x20 in ~0.2 s.

This looks like a panel firmware race (a command arriving while the panel commits an arm/alarm transition wedges the session), but the library can avoid it by not sending the fault check at that exact moment. PR to follow.

michael-z-liu:
Fixes #68.

Sending `REQUEST_PANEL_SYSTEM_STATUS (0x20)` in the same millisecond as an arm/alarm history event can wedge the automation session on Solution 3000 panels, leading to a command-skew reset and ~40 s of unavailability. This defers the fault check by 5 s and skips it if the connection dropped in the meantime.

Tested on a Solution 3000 (fw 2.1.3, B426-M 03.16.202): before, 4 of 5 arms reset the connection; after, 5 of 5 arms and 5 of 5 disarms were clean. The only side effect is fault status updating up to 5 s later after a history event. Happy to restrict this to Solution/AMAX panels if you prefer.


mag1024:
Hi -- thanks for the PR!
We already have some delaying around arm transitions for in the `_area_on_off_finalizer`.
Can we generalize this somehow? Maybe a blackout around the arming transition?

michael-z-liu:
Hi @mag1024, thanks for the quick reply, and good to e-meet you.

That's exactly where I ended up. I've pushed a new commit that replaces the fault-check delay with a general blackout in `_send_command`, keyed off both area status changes and history events, so it also covers the alarm detail lookup, output status reloads, and commands issued by users and automations. Details and test results are in the updated description.

Good catch on `_area_on_off_finalizer` - I came across that 30 s delay and its comment while digging, and it looks like the same class of problem.

Happy to adjust the blackout duration or limit it to Solution/AMAX if you'd prefer.

P.S. This PR is heavily AI-assisted: the diagnosis came from reading debug logs together with Claude, and the patch was written that way too. Everything in it has been verified on my own Solution 3000, but a third eye on the code itself would be much appreciated, particularly on whether the blackout belongs in _send_command or somewhere better.

